PR #2, opened 2015-07-22
Add CMakeLists.txt

proller:


tuexen:
Is there a chance that you also provide a CMAKE file to build the examples in programs/
That would be awesome...

Michael

proller:
builds two versions of lib:  .so and .a 
supports install of .so and includes
added usrsctp.pc.in for pkg-config support
(now package for many os can be easily created)
some parts taken from https://github.com/msgpack/msgpack-c/blob/master/CMakeLists.txt

tuexen:
Great. Let me test it... I'm currently at the IETF meeting, so it might take a couple of days.

tuexen:
Felix took your patch, improved it and tested it under Linux, Mac OS X, FreeBSD and Windows.
The resulting pull request was #13 and got merged. Thanks for providing the initial patch!
Please let us know if you find any issue with the cmake support. The pkg-config support will be added as a separate commit.

PR #11, opened 2015-08-23
FIX: ekr_loop is not built on windows

weinrank:


tuexen:
ekr_loop doesn't run on windows, so why build it?

weinrank:
My fault...

PR #12, opened 2015-08-26
Programs improvements

weinrank:
Let ekr_loop exit on errors and stop the ekr_* execution if not enough arguments have been supplied.


tuexen:
Can you use if () instead of if() and update the patch?

weinrank:
@tuexen Done!

tuexen:
Taken. Thanks.

PR #17, opened 2015-09-23
Fix clang build.

ossy-szeged:


mcatanzaro:
Well using -Werror for release builds is the real mistake... that never makes sense, since it guarantees you'll be broken next time a compiler gets a new or improved warning.

My recommendation is to combine AX_IS_RELEASE with AX_COMPILER_FLAGS in order to use -Werror only for development builds, so users don't get bit by it.

ossy-szeged:
I can't find AX_IS_RELEASE or AX_COMPILER_FLAGS in this project.
Where should I search them and how to fix the build properly?

mcatanzaro:
They're in autoconf-archive. You can add an m4 folder and copy them into there, but most GNOME modules are just requiring people to have autoconf-archive installed as a dependency.

Documentation:

https://www.gnu.org/software/autoconf-archive/ax_is_release.html
https://www.gnu.org/software/autoconf-archive/ax_compiler_flags.html

Usage examples:

https://git.gnome.org/browse/epiphany/commit/?id=16116312f12cd62c056ad2d7c95f31bc4405d314
https://git.gnome.org/browse/epiphany/commit/?id=d9a7d8e542357a321a49de4ce2b69855413bfd11 (but note the typo "-Werror is not enabled for builds from git", that should be "now" and not "not")

The point is to let the project developers get -Werror, without screwing over people trying to build the project with a different set of warnings.

mcatanzaro:
You should probably wait to see if the devs want that before adding it, though. They might have another preference for conditionally enabling -Werror, or they might prefer to simply remove it.

tuexen:
Interesting... We use in
http://212.201.121.77:18010/waterfall
several slaves with clang (FreeBSD amd64, FreeBSD arm, OS X) as the compiler and don't see the issue. Any idea why?

Looking at:
if (((&stcb->asoc.send_queue)->tqh_first == ((void*)0))) {
The code comes from
if (TAILQ_EMPTY(&stcb->asoc.send_queue)) {
which is a normal pattern we use...

ossy-szeged:
I have no idea why your clang builders didn't notice this issue.

But it's easy to reproduce it on Linux. If you set CC=clang and CXX=clang++ ,
just call ./bootstrap && ./configure && make

ossy-szeged:
Thanks for the autoconf infos, but unfortunately I have to say that I don't have any time for this issue.

The proposed fix works, it's more than enough for me. If you would like to fix it in a 
different way, feel free to do it yourself, but I won't have any time for helping to do it.

tuexen:
I removed the SCTP_ENABLE_UDP_CSUM() macro, since it is not need in the userland stack. It is only useful in the FreeBSD kernel. See https://github.com/sctplab/usrsctp/commit/adcf33be601fed0f741eb458394cab984901c91a. I also added a --enable-warnings-as-errors / --disable-warnings-as-errors option to the configure script in
https://github.com/sctplab/usrsctp/commit/078ff3252f73327e0ac11d6fd5eff62011f6646e. The default is on. I can change that to off, if you prefer. Does this address your issue?

ossy-szeged:
Thanks, these commits fixes everything what we needed.

PR #22, opened 2015-10-10
Data racing fix on reading ticks in sctp_callout.

h8liu:
Caught by a data race detector.


tuexen:
I don't think this is really an issue, but I'll merge it to silence the warnings. Some tweaking needs to be applied to get it working on other platforms.

PR #19, opened 2015-09-26
Fix compilation with GCC 5

saghul:
GCC 5 only allows the use of `__FUNCTION__` as an extension. so
it must be prefixed by `__extension__`, though the standard (since C99)
`__func__` is preferred. So use the latter since -std=c99 is required anyway.

Fixes the following errors:

~~~~
In file included from ./netinet/sctp_os.h:72:0,
                 from netinet/sctp_auth.c:38:
netinet/sctp_auth.c: In function 'sctp_auth_key_acquire':
netinet/sctp_auth.c:563:4: error: ISO C does not support '__FUNCTION__' predefined identifier [-Werror=pedantic]
    __FUNCTION__, (void *)stcb, key_id, skey->refcount);
    ^
./netinet/sctp_os_userspace.h:560:31: note: in definition of macro 'SCTP_PRINTF'
   SCTP_BASE_VAR(debug_printf)(__VA_ARGS__); \
                               ^
netinet/sctp_auth.c:561:3: note: in expansion of macro 'SCTPDBG'
   SCTPDBG(SCTP_DEBUG_AUTH2,
   ^
netinet/sctp_auth.c: In function 'sctp_auth_key_release':
netinet/sctp_auth.c:583:4: error: ISO C does not support '__FUNCTION__' predefined identifier [-Werror=pedantic]
    __FUNCTION__, (void *)stcb, key_id, skey->refcount);
    ^
./netinet/sctp_os_userspace.h:560:31: note: in definition of macro 'SCTP_PRINTF'
   SCTP_BASE_VAR(debug_printf)(__VA_ARGS__); \
                               ^
netinet/sctp_auth.c:581:3: note: in expansion of macro 'SCTPDBG'
   SCTPDBG(SCTP_DEBUG_AUTH2,
   ^
netinet/sctp_auth.c:592:5: error: ISO C does not support '__FUNCTION__' predefined identifier [-Werror=pedantic]
     __FUNCTION__, (void *)stcb, key_id, skey->refcount);
     ^
./netinet/sctp_os_userspace.h:560:31: note: in definition of macro 'SCTP_PRINTF'
   SCTP_BASE_VAR(debug_printf)(__VA_ARGS__); \
                               ^
netinet/sctp_auth.c:590:4: note: in expansion of macro 'SCTPDBG'
    SCTPDBG(SCTP_DEBUG_AUTH2,
    ^
cc1: all warnings being treated as errors
~~~~

PR #23, opened 2015-10-17
GCC5 Support

mattvv:
gcc-5 adjusted the gcc interpretation of pedantic, which usrsctp violates.

Temporarily removed the pedantic flag until this can be resolved.


tuexen:
Can you provide a list of warnings gcc5 generates?

tuexen:
Issue #19 has been fixed. In case you have more problems, please report them. In general, I would suggest to use --disable-warnings-as-errors as a parameter to the configure script instead of disabling the warnings. Therefore closing it without merging it.

PR #26, opened 2015-10-27
Cmake: fix 'make install' and use dynamic linking by default for UNIX

weinrank:


usrsctplib/CMakeLists.txt
```diff
@@ -275,7 +275,9 @@ ENDIF ()
 
 INSTALL (TARGETS usrsctp usrsctp-static DESTINATION ${CMAKE_INSTALL_LIBDIR})
 #INSTALL (DIRECTORY include DESTINATION ${CMAKE_INSTALL_PREFIX})
-INSTALL (FILES ${usrsctp_root_HEADERS} DESTINATION include/usrsctp)
-INSTALL (FILES ${usrsctp_netinet_HEADERS} DESTINATION include/usrsctp/netinet)
-INSTALL (FILES ${usrsctp_netinet6_HEADERS} DESTINATION include/usrsctp/netinet6)
-INSTALL (FILES ${CMAKE_CURRENT_BINARY_DIR}/usrsctp.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
+#INSTALL (FILES ${usrsctp_root_HEADERS} DESTINATION include/usrsctp)
+#INSTALL (FILES ${usrsctp_netinet_HEADERS} DESTINATION include/usrsctp/netinet)
```
tuexen: Why is this line needed? What are you installing there?

usrsctplib/CMakeLists.txt
```diff
@@ -275,7 +275,9 @@ ENDIF ()
...
+#INSTALL (FILES ${usrsctp_netinet6_HEADERS} DESTINATION include/usrsctp/netinet6)
```
tuexen: Why is this line needed? What are you installing there?

usrsctplib/CMakeLists.txt
```diff
@@ -275,7 +275,9 @@ ENDIF ()
 
 INSTALL (TARGETS usrsctp usrsctp-static DESTINATION ${CMAKE_INSTALL_LIBDIR})
 #INSTALL (DIRECTORY include DESTINATION ${CMAKE_INSTALL_PREFIX})
-INSTALL (FILES ${usrsctp_root_HEADERS} DESTINATION include/usrsctp)
-INSTALL (FILES ${usrsctp_netinet_HEADERS} DESTINATION include/usrsctp/netinet)
-INSTALL (FILES ${usrsctp_netinet6_HEADERS} DESTINATION include/usrsctp/netinet6)
-INSTALL (FILES ${CMAKE_CURRENT_BINARY_DIR}/usrsctp.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
+#INSTALL (FILES ${usrsctp_root_HEADERS} DESTINATION include/usrsctp)
```
tuexen: Why do you want to create a directory to install a single file? Is that the normal way? Just wondering...

weinrank: I don't - that's why i commented it out. 
This line is from the initial CMAKE commit

usrsctplib/CMakeLists.txt
weinrank: Currently all header-files from netinet are copied to this directory.
I don't think this is necessary...

usrsctplib/CMakeLists.txt
weinrank: Like in the line above the (single) header file from netinet6 is copied to this directory.

usrsctplib/CMakeLists.txt
tuexen: Missed that it is commented out. Why not remove it?

usrsctplib/CMakeLists.txt
tuexen: Don't install any header file except usrsctp.h. Only that is the interface.

usrsctplib/CMakeLists.txt
tuexen: See above...

usrsctplib/CMakeLists.txt
weinrank: This is exactly what we achieve by this patch.
The behavior before this patch is to copy all files to <includedir>/usrsctp/

With this patch *only* the usrsctp.h is copied  -  and not to <includedir>/usrsctp/ but to <includedir>/ - this is how the traditional method does.

usrsctplib/CMakeLists.txt
weinrank: Experimental technology - so it's easier to track changes to the initial version.

tuexen: Since we don't want to do that take it out.
If you want to have some historical stuff. Commit it as you got it and then take the lines out...

weinrank:
@tuexen Cleaned the cmake file and removed old .cvsignore

Issue #1, opened 2015-07-22
Can I block a connection or an association when using callback?


swordfeng:
I want to prevent the peer from writing more data to this socket. According to RFC4960 I need to tell the peer that my rwnd is 0.
I can stop calling recv() when in non-callback mode. But when using a callback, I can't stop the calling of the callback. Are there any APIs can help?

stewrtrs:
The rwnd is something that is controlled by the actual SCTP implementation. You are
correct in that if a application *stops* reading (calling recv) then the implementation
will buffer up to the rwnd bytes of data (waiting for the user to read it) and then finally
start to send a rwnd of 0.  From the time you stop calling recv() to the time it stops
would vary depending on how much data the peer was sending you since the peer
must fill the rwnd before the rwnd == 0 condition would occur.

The callback API has no way to stop since it has already performed the read for you and
is handing you the data.

In some past implementation’s of such an API I have heard where folks would have the
callback not include the data, but instead say “You have data to read” and then you would
have to do a recv() call to acquire the data. Since it was not implemented that way then
I don’t think you can do anything to stop the data with current mechanisms.

You could, of course, always create an application to application message that says “Stop please
don’t send me data”. As long as your peer honored that then you would not receive more..
That is probably your only choice with the callback API.

Best wishes

R
On Jul 22, 2015, at 7:33 AM, Jianfeng Zhang <notifications@github.com> wrote:

> I want to prevent the peer from writing more data to this socket. According to RFC4960 I need to tell the peer that my rwnd is 0.
> I can stop calling recv() when in non-callback mode. But when using a callback, I can't stop the calling of the callback. Are there any APIs can help?
> 
> —
> Reply to this email directly or view it on GitHub.
> 

--------
Randall Stewart
rrs@netflix.com
803-317-4952

PR #32, opened 2015-11-17
Put the stream in SCTP_STREAM_RESET_PENDING for both send_in and send_out

shiretu:


tuexen:
What is the bug you want to fix?

shiretu:
I get `ENOENT` when trying to reset streams, although I clearly see the stream being alive, because I just received data over it (the ACK from setting the label on wrtc channel). I was performing some tests where the channel is immediately closed after getting the that ACK. 

Also, the fix is not working by itself. I have to call setsockopt with RESET_STREAMS from the dry event.

shiretu:
Basically, that is a 2 fold:

1. `stcb->asoc.strmout[i].state == SCTP_STREAM_RESET_PENDING`: is satisfied by the patch I have submited
2. `(TAILQ_EMPTY(&stcb->asoc.strmout[i].outqueue))` is satisfied by calling reset on streams from dry event

tuexen:
I still don't understand which state you are running into and what is wrong with it. SCTP_STREAM_RESET_PENDING is related to outgoing streams. Receiving data is related to incoming streams. Incoming and outgoing streams are handled independently.  WebRTC uses an outgoing stream and an incoming stream with the same identifier as a data channel, but in the SCTP layer they are handled independently.

I tested the resetting of outgoing streams, since they are used for WebRTC.
So which system calls are you doing in which sequence to observe a behaviour which you don't expect. I need to understand that to figure out what the underlying problem is and what an appropriate fix is.

shiretu:
I open and close channels in a very rapid succession. For example, supposed we do the following sequence:

1. Open a channel (SID: 1)
2. Close the channel (SID: 1) by doing reset
3. Wait for both reset ACK to arrive (incoming and outgoing)
4. At this point, SID 1 should be available to be opened again
5. Open a channel again (SID: 1, since now is free)
6. Try to reset it again (SID: 1). At this point, I get ENOENT, although I did received data over it so it is clearly in there.

Root problem here is that setsockopt with stream reset, sometimes, fails to set the internal state of the stream to `SCTP_STREAM_RESET_PENDING`. Later in the call chain, inside `sctp_add_stream_reset_out` function, `stcb->asoc.strmout[i].state == SCTP_STREAM_RESET_PENDING` is not satisfied, giving me ENOENT at the return from setsockopt

tuexen:
OK. Let me look into this. Need some time, since I'm currently debugging another problem...

tuexen:
Your fix is incorrect. Only outgoing streams should only be marked pending. It does not matter what is requested about incoming streams. Once and outgoing stream which is pending is dry, the stream reset is sent out. This means that before resetting a stream, the SCTP stack will ensure that all user data for that stream is received by the peer.
Haven't figured out your ENOENT problem yet.

shiretu:
More information:

I have added a simple logging macro than used it in here:
https://github.com/sctplab/usrsctp/blob/master/usrsctplib/netinet/sctp_output.c#L12203

Like this:

```
static int
sctp_add_stream_reset_out(struct sctp_tcb *stcb, struct sctp_tmit_chunk *chk,
                          uint32_t seq, uint32_t resp_seq, uint32_t last_sent)
{
	uint16_t len, old_len, i;
	struct sctp_stream_reset_out_request *req_out;
	struct sctp_chunkhdr *ch;
	int at;
	int number_entries=0;

	ch = mtod(chk->data, struct sctp_chunkhdr *);
	old_len = len = SCTP_SIZE32(ntohs(ch->chunk_length));
	/* get to new offset for the param. */
	req_out = (struct sctp_stream_reset_out_request *)((caddr_t)ch + len);
	/* now how long will this param be? */
	for (i = 0; i <stcb->asoc.streamoutcnt; i++) {
		if ((stcb->asoc.strmout[i].state == SCTP_STREAM_RESET_PENDING) &&
		    (TAILQ_EMPTY(&stcb->asoc.strmout[i].outqueue))) {
			number_entries++;
		}
	}
	if (number_entries == 0) {
		MY_LOG("Previous loop did not incremented number_entries. Dumping:");
		for (i = 0; i <stcb->asoc.streamoutcnt; i++) {
			MY_LOG("i: %d; stcb->asoc.strmout[i].state: %d; TAILQ_EMPTY(&stcb->asoc.strmout[i].outqueue): %d",
					i,stcb->asoc.strmout[i].state,TAILQ_EMPTY(&stcb->asoc.strmout[i].outqueue));
		}
		MY_LOG("Dumping complete. Doing return(0)");
		return(0);
	}
```

The output is as follows:
http://pastebin.com/raw.php?i=hvfGUWfb

shiretu:
Indeed, my patch is faulty. But ENOENT is caused by the fact that the output queue is not empty for the closed streams. They are not empty because the close occurred in a very rapid succession:

1. Open stream
2. Close stream
3. Open Stream (same side as the one at #1)
4. Close Stream --> here we get ENOENT

tuexen:
> On 21 Nov 2015, at 09:15, Gavriloaie Eugen-Andrei <notifications@github.com> wrote:
> 
> Indeed, my patch is faulty. But ENOENT is caused by the fact that the output queue is not empty for the closed streams. They are not empty because the close occurred in a very rapid succession:
> 
> 	• Open stream
> 	• Close stream
> 	• Open Stream (same side as the one at #1)
> 	• Close Stream --> here we get ENOENT
OK. I looked at the code and understand why you get ENOENT.
This error should not be reported you since the request will
be handled by the SCTP stack later. So it is OK to have it
*within* the SCTP stack, but it should not be reported to the
user.

While looking at the code I also found a bug in the condition
to move from SCTP_STREAM_RESET_PENDING to SCTP_STREAM_RESET_IN_FLIGHT.
We are not considering the chunks of the stream under consideration which
are in the send_queue or sent_queue.

Let me figure out how to fix and test this. I'll send you an e-mail
when I have fixed it. Then you can test the fix.

Best regards
Michael

tuexen:
> On 21 Nov 2015, at 09:15, Gavriloaie Eugen-Andrei <notifications@github.com> wrote:
> 
> Indeed, my patch is faulty. But ENOENT is caused by the fact that the output queue is not empty for the closed streams. They are not empty because the close occurred in a very rapid succession:
> 
> 	• Open stream
> 	• Close stream
> 	• Open Stream (same side as the one at #1)
> 	• Close Stream --> here we get ENOENT
The ENOENT problem should be fixed by
https://github.com/sctplab/usrsctp/commit/e5bd44707bf3edc47f4b74336452cfd3d647dfe2

I'm still working on a patch which ensures that an outgoing stream is
only reset after all pending messages have been acknowledged by the
peer. We are currently testing it...

Best regards
Michael

tuexen:
I think the problem is solved...

PR #38, opened 2015-11-28
Single-threaded operation, additional platform support, vectored send

ccaughie:
This is a series of patches based on some modifications that we have been using in my organization. There are three main changes:

1) Allow usrsctp to be driven entirely from the application's own event loop, which may be processing other sockets in addition to the UDP sockets that encapsulate the SCTP traffic.
2) Add usrsctp_sendvec to support vectored sends (I haven't added vectored receives only because I have not needed them yet).
3) Add support for other compilers/platforms, in particular MinGW (however I have not added MinGW support to the makefiles since we use an external build utility to build the library).

Note that the last change is essential particularly for MinGW usage; it allows us to force the use of Winsock error codes over any others that might have been picked up from an errno.h file. It's essential for correct operation that all usrsctp modules and all application modules are using the same error code definitions, otherwise the application can't recognize important codes such as EWOULDBLOCK.

### DIFF
--- a/usrsctplib/netinet/sctp_usrreq.c
+++ b/usrsctplib/netinet/sctp_usrreq.c
@@ -34,7 +34,7 @@
 
 #ifdef __FreeBSD__
 #include <sys/cdefs.h>
-__FBSDID("$FreeBSD: head/sys/netinet/sctp_usrreq.c 359405 2020-03-28 20:25:45Z tuexen $");
+__FBSDID("$FreeBSD: head/sys/netinet/sctp_usrreq.c 359410 2020-03-28 22:35:04Z tuexen $");
 #endif
 
 #include <netinet/sctp_os.h>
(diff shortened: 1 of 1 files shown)


tuexen:
Thanks for providing your patch. I'll integrate some of the stuff in separate commit and will add comments to other things. However, it might take some time...

gvanem:
I've *not* tried your PR, but `gettimeofday()` should be for `_MSC_VER` only:
```
--- a/tsctp.c 2016-04-04 09:22:18
+++ b/tsctp.c 2016-04-04 09:26:34
@@ -117,7 +117,7 @@
        done = 1;
 }

-#ifdef _WIN32
+#ifdef _MSC_VER
 static void
 gettimeofday(struct timeval *tv, void *ignore)
 {
```

Since `gettimeofday()` has been in MinGW since dawn of time.

gfodor:
Any info on the status of this PR?

tuexen:
This PR combines multiple independent things. I think I took what can be considered bug fixes. MinGW support is still on the ToDo list. We have set up an MinGW buildbot, but the errors shown there and what is reported differ. I have not included the additional iov-based API calls. Support for external event loop is being worked on in the NEAT fork or usrsctp to get some experience. This will be merged eventually. The current focus is on bugfixing (issues detected by coverity) and and fixing accounting bug for SCTP-IDATA. Any particular issue in this PR you are interested in?

gfodor:
Ah thanks for the info. My main need is external event loop support, as I'm investigating the viability of integrating this library with erlang/OTP to enable communicating with webrtc data channels. Are there issues with the implementation on this PR that I could perhaps help address? It seems relatively straightforward but I am new to this project.

tuexen:
Ahh, I see. That seems interesting. The person who did this work for the `libuv` event loop in https://github.com/NEAT-project/usrsctp-neat is currently on vacation. Once she is back, I can see if we can integrate merge this stuff. Need to review it to make sure the API is generic enough. In the meantime you can have a look. The `libuv` project where it is used is the leat library.

lgrahl:
@gfodor No need to wait, you can use [usrsctp-neat](https://github.com/NEAT-project/usrsctp-neat) which will be merged back eventually. It works in the following way:

1. Set the [upcall handler](https://github.com/rawrtc/rawrtc/blob/700158fab22d17b7b1eb9c7219aff986cf19ff63/src/librawrtc/sctp_transport.c#L2113) and then you can check the events (read, write, error) [in the handler](https://github.com/rawrtc/rawrtc/blob/700158fab22d17b7b1eb9c7219aff986cf19ff63/src/librawrtc/sctp_transport.c#L1758:L1803).
2. You'll also need to call `usrsctp_handle_timers` every ~10ms.

However, instead of rolling your own WebRTC stack you could also use [RAWRTC](https://github.com/rawrtc/rawrtc) and create bindings for your preferred language. That would be pretty neat and I'm happy to help! :smiley: If you have any questions, feel free to ask them on our [gitter](https://gitter.im/rawrtc/Lobby?source=orgpage) channel (or contact me via mail).

gfodor:
That's great, thanks all! Will look into RAWRTC as well.

falkevik:
@gfodor seems like we are after the same thing here. I have integrated usrsctp with erlang/OTP. My first approach was to use pipes that I could put into the select as for other fds in inet_drv.c. But then I switched to event callbacks which will schedule the concerned port for the socket.
If you are interested please have a look at https://github.com/falkevik/otp/tree/usrsctp_event_cb_otp_upgrade

jmillan:
Hi,

What's the status of this PR?

jmillan:
Hi @ccaughie,

> 1. Allow usrsctp to be driven entirely from the application's own event loop, which may be processing other sockets in addition to the UDP sockets that encapsulate the SCTP traffic.

May I ask for the exact commit where you added this feature? I would like to have *only* those changes.

I imagine [this](https://github.com/sctplab/usrsctp/pull/38/commits/010fcbc6f193152607dc267c7e7e573c9bffc111) is the commit I look for. I would like to know if this is the only needed commit, and whether it contains changes not relevant to point 1.

Thank you.

PR #39, opened 2015-12-04
Fix calling convention mismatch for thread creation argument

rnk:
The Windows CreateThread API expects a function that is __stdcall.
Passing it a normal function pointer that is __cdecl causes crashes in
some situations.

There are more instances of this bug in programs/, but they pass
arguments and so they are not easy to adapt through sctp_thread_create.

More context: http://crbug.com/562308


tuexen:
Thanks for reporting the issue and providing the fix. In a followup commit I just renamed sctp_create_thread to sctp_userspace_create_thread and reverted a whitespace change you provided by accident. See
https://github.com/sctplab/usrsctp/commit/7a1a2fa16d55a343ea8605977ba06d553155b328
Will fix the examples in programs...

tuexen:
.. and you need https://github.com/sctplab/usrsctp/commit/035d537a97b31d14027eaaa7aa46d8c1dadc1eb6 to get it compiling and running on Windows... Buildbots are green now.

rnk:
Thanks!

PR #44, opened 2016-01-04
use prctl instead of (glibc-specific) pthread_setname_np on Linux

mostynb:
With this small change, this code should build with other libc's on linux (eg uClibc).


tuexen:
Merged. Thanks a lot for providing the patch.

PR #48, opened 2016-02-04
Fix memory leak getting MTU under Windows

jswirl:
Free pAdapterAddrs after use.


tuexen:
Fixed in a slightly different way in https://github.com/sctplab/usrsctp/commit/36970fc3e8400dde4ea32a779488a14bfb9db7de.
Thanks for reporting the issue and providing a fix.

PR #70, opened 2016-04-19
New sample: http_client

weinrank:


programs/http_client.c
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright (C) 2015 Felix Weinrank
```
tuexen: Living in the past?

programs/http_client.c
```diff
@@ -0,0 +1,186 @@
...
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+/*
+ * Usage: http_client remote_addr remote_port [local_port] [local_encaps_port] [remote_encaps_port]
+ */
```
tuexen: Would it make sense to add the object which is requested as a parameter?
That way we could request `cgi-bin/he` to get the transport protocol choice information...

programs/http_client.c
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+	while (!done) {
+#ifdef _WIN32
+		Sleep(1*1000);
+#else
+		sleep(1);
+#endif
+	}
+	usrsctp_get_stat(&stat);
```
tuexen: Not sure this is useful here...

programs/http_client.c
weinrank: feature added in a147886

programs/http_client.c
weinrank: typo fixed in a147886

programs/http_client.c
weinrank: removed in a147886

PR #87, opened 2016-06-16
Win ipv6 fix

weinrank:
This PR fixes an issue with IPV6 for Windows.
Windows returns INVALID_SOCKET instead of -1 if an error occurs - Socket is unsigned in windowsland...
https://msdn.microsoft.com/en-us/library/windows/desktop/ms740516(v=vs.85).aspx

The first commit (7a2b27e) fixes the problem.
The second (4d6dd67) one defines INVALID_SOCKET as -1 for platforms where INVALID_SOCKET is not defined by default to have a consistent error handling.

### DIFF
--- a/usrsctplib/netinet/sctp_pcb.h
+++ b/usrsctplib/netinet/sctp_pcb.h
@@ -317,6 +317,11 @@ struct sctp_base_info {
 	userland_thread_t recvthreadroute;
 #endif
 #endif
+
+#ifndef INVALID_SOCKET
+#define INVALID_SOCKET -1
+#endif
+
 #ifdef INET
 #if defined(__Userspace_os_Windows)
 	SOCKET userspace_rawsctp;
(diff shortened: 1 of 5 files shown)


usrsctplib/netinet/sctp_pcb.h
```diff
@@ -317,6 +317,11 @@ struct sctp_base_info {
 	userland_thread_t recvthreadroute;
 #endif
 #endif
+
+#ifndef INVALID_SOCKET
+#define INVALID_SOCKET -1
+#endif
+
```
tuexen: Can we move that definition to `sctp_os_userspace.h`? We can also be specific and define it whenever we are no on Windows.

weinrank:
*up*

tuexen:
Closing as discussed with Felix. The issue is fixed in https://github.com/sctplab/usrsctp/commit/631705df46dc16927340f847ff16022cdb7a8312.

PR #93, opened 2016-07-19
Fixing heap overflow in sctp_send_deferred_reset_response.

taylor-b:
sctp_add_stream_reset_result was being called before chunk_length
was set, causing the result to be written to an undefined location
in memory.


tuexen:
Hi Taylor,

thanks for the bug report and the provided patch. I merged it and also merged it to the FreeBSD kernel.

Best regards
Michael

PR #94, opened 2016-07-22
Fix Windows compile error when neither INET nor INET6 is defined.

taylor-b:


tuexen:
Thanks. Merged. We will test all combination of INET/INET6 support shortly. I actually thought that the buildbots will do this already, but that is not the case.

taylor-b:
That's good to hear. Thanks!

PR #100, opened 2016-09-09
Fixed the QMD_TAILQ_CHECK_TAIL invariant check.

skvlad:
The invariant check was incorrect, and could never succeed for a
non-empty queue. I have discovered this when I enabled invariant checks while debugging a timer queue corruption. 

The fix for the queue corruption is in a separate pull request.


skvlad:
I misunderstood the logic of the function. It was failing in my tests as the queue was already corrupted by that time.
The check is correct as is.

PR #101, opened 2016-09-09
Fixed a race condition causing timer queue corruption.

skvlad:
The SCTP_OS_TIMER_DEACTIVATE macro was not locking the timer queue while
it modified the flags. If the timer fires as the read-modify-write
operation is happening, the timer will be reinserted into the queue
without the SCTP_CALLOUT_PENDING flag set. When the timer fires
subsequently, it gets added to the queue for the second time and its
prev/next pointers get overwritten - which can create a loop in the=
timer queue.

The fix is to lock the queue before modifying timer flags.

### DIFF
--- a/usrsctplib/netinet/sctp_callout.h
+++ b/usrsctplib/netinet/sctp_callout.h
@@ -96,7 +96,11 @@ int sctp_os_timer_stop(sctp_os_timer_t *);
 #define SCTP_OS_TIMER_STOP_DRAIN SCTP_OS_TIMER_STOP
 #define	SCTP_OS_TIMER_PENDING(tmr) ((tmr)->c_flags & SCTP_CALLOUT_PENDING)
 #define	SCTP_OS_TIMER_ACTIVE(tmr) ((tmr)->c_flags & SCTP_CALLOUT_ACTIVE)
-#define	SCTP_OS_TIMER_DEACTIVATE(tmr) ((tmr)->c_flags &= ~SCTP_CALLOUT_ACTIVE)
+#define	SCTP_OS_TIMER_DEACTIVATE(tmr) do {				\
+	SCTP_TIMERQ_LOCK();						\
+	(tmr)->c_flags &= ~SCTP_CALLOUT_ACTIVE;				\
+	SCTP_TIMERQ_UNLOCK();						\
+} while (0)
 
 #if defined(__Userspace__)
 void sctp_start_timer(void);


tuexen:
Could you explain which two threads would be involved in the race condition?
SCTP_OS_TIMER_DEACTIVATE is only called from the function `sctp_timeout_handler()`,
which runs in the timer thread. So how can "read-modify-write operation is happening"
and "the timer fires" happen in a race?

skvlad:
I was wrong in the assumption the race happened the way it's described above. Turns out it is slightly different.

In the specific circumstances I saw this happen, the timer involved in the race is SCTP_TIMER_TYPE_ADDR_WQ. You are indeed correct that SCTP_OS_TIMER_DEACTIVATE is called from the timer thread - however that thread is not the only one that can change the timer. The user application thread can do that too - in particular, when calling usrsctp_register_address / usrsctp_deregister_address.

Here is the smallest repro I found that crashes when INVARIANTS are enabled (for me at least), or gets stuck in an infinite loop when they are off:

~~~~
usrsctp_init(0, &OnSctpOutboundPacket, &DebugSctpPrintf);
char *ptr = new char;
for (int i = 0; i < 100000; i++) {
  usrsctp_register_address(ptr);
  usrsctp_deregister_address(ptr);
}
~~~~

In my understanding here's what is happening: the usrsctp_deregister_address call schedules a timer 1 tick into the future. While the timer callback is running on the timer thread, it releases the lock - meanwhile the user's application calls usrsctp_deregister_address again and reschedules the timer. 

So the two threads involved in the race are the application thread running sctp_timer_start, and the SCTP timer thread running the callback - both for the same timer.

tuexen:
On which OS are you are to reproduce the issue?

skvlad:
I was testing it on Linux x86_64, building with clang 4.0.0 as part of WebRTC.

tuexen:
I put a test program in the repo (https://github.com/sctplab/usrsctp/blob/master/programs/test_timer.c) which can be used to reproduce the issue on Linux systems. The problem should be limited to the `SCTP_TIMER_TYPE_ADDR_WQ` timer, since there is no lock protecting that timer. Using a global lock for all timers fixes the bug but is too restricting. Will need to figure out which lock will be used for that timer...

tuexen:
This issue should be fixed by https://github.com/sctplab/usrsctp/commit/96939f134c9d643c3550518a692da6a5ca08822b.

tuexen:
Closing this now. If the problem still exists, please reopen.

PR #102, opened 2016-09-19
add usrsctp prefix to mbuf_init()

alfredh:
my program is using both usrsctp and libre SIP stack (https://github.com/creytiv/re).
both libraries define the symbol `mbuf_init` and when building, the linker fails
with duplicated symbols.

To avoid this, I suggest that usrsctp adds the `usrsctp` prefix to the mbuf_init function.

### DIFF
--- a/usrsctplib/netinet/sctp_pcb.c
+++ b/usrsctplib/netinet/sctp_pcb.c
@@ -6844,7 +6844,7 @@ sctp_pcb_init()
 	TAILQ_INIT(&SCTP_BASE_INFO(callqueue));
 #endif
 #if defined(__Userspace__)
-	mbuf_init(NULL);
+	usrsctp_mbuf_init(NULL);
 	atomic_init();
 #if defined(INET) || defined(INET6)
 	recv_thread_init();
--- a/usrsctplib/user_mbuf.c
+++ b/usrsctplib/user_mbuf.c
@@ -328,11 +328,11 @@ m_tag_setup(struct m_tag *t, u_int32_t cookie, int type, int len)
 /************ End functions to substitute umem_cache_alloc and umem_cache_free **************/
 
 /* __Userspace__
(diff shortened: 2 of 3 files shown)


tuexen:
We use the `usrsctp` prefix for symbols which are exported. `mbuf_init()` shouldn't be called from anyone outside and is not considered part of the exported API. Therefore this prefix doesn't seem appropriate.
Which OS are you using? Possibly we could limit the visibility of exported symbols.

alfredh:
we are using the following platforms:

* OSX
* Linux
* iOS
* Android

the compiled usrsctp library is a static .a file (libusrsctp.a) which contains a
flat table of all symbols, include `mbuf_init` which is marked with `T`.

if your mbuf_init was static it would work, as it would have been marked `t`.

alfredh:
I have found a workaround for my application. should I close this PR ?

lgrahl:
Although I've found a workaround as well, I believe this is still a valid issue. I don't know if you've encountered the same problem @alfredh but when (accidentally) using usrsctp as a dynamic library, it gets really really messed up (this might be machine- and cmake-specific though, I can't really explain the problem properly, so if you're asking yourself wth I'm talking about, I can provide a screenshot).

tuexen:
I could change `mbuf_init()` to something like `mbuf_initialize()`, to avoid the name conflict.
@lgrahl Would that solve your problem?

lgrahl:
Yeah, that should solve our problem.

alfredh:
@tuexen thanks that would also solve my problem :)

tuexen:
Will fix it...

tuexen:
Fixed in https://github.com/sctplab/usrsctp/commit/b018ce66a4f9323c788619299c291a8d90854355.

alfredh:
many thanks Michael :)

I tested latest version from HEAD and it works fine.

PR #112, opened 2016-11-30
Debug version must be compiled with "-g".

dreibh:
Debug version must be compiled with "-g". Otherwise, gdb will have no symbol files for debugging.


weinrank:
If `CMAKE_BUILD_TYPE` is set to `DEBUG`, cmakes adds this option automatically.

tuexen:
Closing since it seems to be not an issue. Please reopen, if the problem actually still exists.

PR #158, opened 2017-07-28
Use '#pragma pack (pop)' instead of '#pragma pack ()'

zmodem:
The code is setting the packing alignment with

`  #pragma pack (push, 1)`

and would then reset it with

`  #pragma pack ()`

The latter directive resets the packing alignment to the default value,
which means if some other packing alignment was in effect before the
'push', that is now lost.

This patch resets the packing alignment with

`  #pragma pack (pop)`

instead, to patch the 'push' operation.

(This problem was found in a Chromium build by a new Clang warning,
-Wpragma-pack.)

### DIFF
--- a/programs/rtcweb.c
+++ b/programs/rtcweb.c
@@ -140,7 +140,7 @@ struct rtcweb_datachannel_ack {
 } SCTP_PACKED;
 
 #ifdef _WIN32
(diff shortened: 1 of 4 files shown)


tuexen:
Thanks a lot for the patch. It has been merged.

PR #159, opened 2017-08-03
Fix one remaining #pragma pack()

zmodem:
Somehow I missed this in my previous patch.


tuexen:
Merged. Thanks for the patch.

PR #162, opened 2017-08-30
callout: assign the next pointer of callqueue in lock

justinjoy:
Otherwise, the next pointer is able to be contaminated while
executing function pointer.

Signed-off-by: Justin Kim <justin.kim@collabora.com>

### DIFF
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -160,10 +160,10 @@ sctp_handle_tick(int delta)
 			c_func = c->c_func;
 			c_arg = c->c_arg;
 			c->c_flags &= ~SCTP_CALLOUT_PENDING;
+			c = sctp_os_timer_next;
 			SCTP_TIMERQ_UNLOCK();
 			c_func(c_arg);
 			SCTP_TIMERQ_LOCK();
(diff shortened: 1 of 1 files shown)


justinjoy:
Although it would be better to explain the symptom with a specific data, I forgot to capture GDB log.
I sent this PR because I have seen very few times that `TAILQ_NEXT(c, tqe)` is same as `c` in [sctp_callout.c#L168](https://github.com/sctplab/usrsctp/blob/master/usrsctplib/netinet/sctp_callout.c#L168) so `SCTP_TIMERQ_UNLOCK` is never called. I suspect that the next pointer of `callqueue` was contaminated while invoking [c_func(c_arg)](https://github.com/sctplab/usrsctp/blob/master/usrsctplib/netinet/sctp_callout.c#L164) with unlocked.

justinjoy:
The bug I've seen is very relevant to #101 because it is reproducible while repeating stop and start stream in a Gstreamer pipeline.

tuexen:
This issue should be fixed by https://github.com/sctplab/usrsctp/commit/96939f134c9d643c3550518a692da6a5ca08822b.

tuexen:
Closing this now. If the problem still exists, please reopen.

PR #173, opened 2017-10-10
Add pthread.h include for os_Fuchsia

jamesr:
Fuchsia is a new operating system (https://fuchsia.googlesource.com/)
that supports pthreads for threading. CMake has support for Fuchsia:

https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/Platform/Fuchsia.cmake

so this wires up the #define.

### DIFF
--- a/usrsctplib/netinet/sctp_os_userspace.h
+++ b/usrsctplib/netinet/sctp_os_userspace.h
@@ -274,7 +274,7 @@ typedef char* caddr_t;
 
 #else /* !defined(Userspace_os_Windows) */
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


tuexen:
Is this based on compile time tests or were you able also to run time testing?

weinrank:
@tuexen 
It seems to be based on compile time tests:
https://github.com/sctplab/usrsctp/blob/master/usrsctplib/CMakeLists.txt#L47

This sets the `__Userspace_os_Fuchsia ` define.

jamesr:
Correct - CMake will set this at build file generation time based on the configuration.

tuexen:
Sorry for not being clear: I was wondering if this only make the usrsctp library compiling or if this patch makes it working on this platform.

jamesr:
Ah yes - that's a much more reasonable question :) This makes the library compile, and by inspection I believe it should work, but I haven't tested that the library works on its own.  Are there tests I could run to verify behavior of this library standalone?

tuexen:
There are examples in the `programs` directory. You can play with `tsctp` or use the standalone program `ekr_loop`.

PR #176, opened 2017-10-13
Add support for being notified by fd

falkevik:
When enabled, one byte will be written on the fd
to notify that the socket have something to be read.

I.e. create a pipe where you can do select(2) poll(2) on one side.
And when there are data ready to be read the fd are read ready.
Along with the usrsctp socket that also can be read.

This is used to hook in usrsctp into erlang / beam network driver
where select(2) or poll(2) are used to multiplex over the sockets.

This is something I use on macos to run erlang with sctp support.
Since it is a work in progress on the erlang beam side.
Maybe a notification for when the socket is ready to be written to could be good as well.

### DIFF
--- a/usrsctplib/netinet/sctp.h
+++ b/usrsctplib/netinet/sctp.h
@@ -139,6 +139,7 @@ struct sctp_paramhdr {
 #define SCTP_NRSACK_SUPPORTED           0x00000030
 #define SCTP_PKTDROP_SUPPORTED          0x00000031
 #define SCTP_MAX_CWND                   0x00000032
+#define SCTP_NOTIF_FD                   0x00000033
 
 /*
  * read-only options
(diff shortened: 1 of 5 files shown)


lgrahl:
You probably want to take a look at [usrsctp-neat](https://github.com/NEAT-project/usrsctp-neat) which has *read*, *write* and *error* event support and will be merged back into this project eventually.

falkevik:
Sounds interesting, could you point me to a commit or similar on how to use the events?
Is there a file descriptor I can do `select(2)` or `poll(2)` on?

lgrahl:
Sure. Basically, you register an event callback function that acts as the
result of a `select` poll.

* Registration of the event callback:
https://github.com/rawrtc/rawrtc/blob/eb3de052a3d7aad2f1f7556a7edd4b8ecb08da74/src/librawrtc/sctp_transport.c#L2113
* The actual event callback function (and its quirks, I've added plenty of
comments to that in the code):
https://github.com/rawrtc/rawrtc/blob/eb3de052a3d7aad2f1f7556a7edd4b8ecb08da74/src/librawrtc/sctp_transport.c#L1761

falkevik:
Thanks @lgrahl I have switched to using events now.
It was a bit hacky to use file descriptors.

PR #181, opened 2017-11-05
receive_cb returning 0 on success

JlnWntr:
Shouldn't this function return 0 on success?
Same would go for http_client.c tsctp.c etc.

### DIFF
--- a/programs/client.c
+++ b/programs/client.c
@@ -279,7 +279,7 @@ receive_cb(struct socket *sock, union sctp_sockstore addr, void *data,
 		}
 		free(data);
 	}
-	return (1);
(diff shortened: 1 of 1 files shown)


lgrahl:
I don't think the return code is being used, see: https://github.com/sctplab/usrsctp/blob/de0630dfeee51edb3a4b6119c9954a89ab87da21/usrsctplib/netinet/sctputil.c#L4856

Not sure why it exists in the first place to be honest.

tuexen:
Yes, the return values is just ignored. Would have been better to declare it as `void (*rcv_cb)()`, but that is too late...

PR #184, opened 2017-11-19
memset-call replacement

JlnWntr:
With this pull request, I suggest using
`  struct STRUCT = {0};`

instead of
   ` memset(&STRUCT, 0 ...);`

### DIFF
--- a/programs/client.c
+++ b/programs/client.c
@@ -297,11 +297,11 @@ main(int argc, char *argv[])
 {
 	struct socket *sock;
 	struct sockaddr *addr, *addrs;
-	struct sockaddr_in addr4;
-	struct sockaddr_in6 addr6;
-	struct sctp_udpencaps encaps;
-	struct sctpstat stat;
-	struct sctp_event event;
+	struct sockaddr_in addr4 = {0};
+	struct sockaddr_in6 addr6 = {0};
+	struct sctp_udpencaps encaps = {0};
+	struct sctpstat stat = {0};
+	struct sctp_event event = {0};
 	uint16_t event_types[] = {SCTP_ASSOC_CHANGE,
 	                          SCTP_PEER_ADDR_CHANGE,
 	                          SCTP_SEND_FAILED_EVENT};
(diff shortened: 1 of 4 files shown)


tuexen:
I don't see any benefit in doing so. All the rest of the code uses `memset()` style initialisation and I think being consistent is a good thing. Sorry.
In general I'm preferring setting variable explicitly before using them instead of initialising them when declaring them.

Best regards
Michael

JlnWntr:
`memset` will tell the compiler to call a function that we expect will set during execution the data in `STRUCT` to zero.

Doesn't `STRUCT={0};` tell the compiler to set the data to zero by itself at compile time?

lgrahl:
In comparison to `memset`, `bla = {0}` can be [easily optimised by the compiler](https://stackoverflow.com/a/10788871).
Note, that it's not required to use `bla = {0}` during declaration. It can be done when initialising later, too.
And another note... it has some [really annoying quirks when using GCC 5.1](https://stackoverflow.com/a/1539162).

While I agree it's *better* to use `{0}`, I doubt it affects performance in a noticeable way, at least for this project.

tuexen:
I don't see how the initialization can be done at compile time in contrast to run time.

stewrtrs:
I don’t comment on these very much.. but I will on this one

The SCTP code is shared between kernel land (FreeBSD/Mac and Windows) and user space. Any move
to not use memset() would have to be in user space code only.. and I really don’t see the advantage
of that. Having a consistent use is far better IMO and far less prone to bugs.

Best wishes

R
> On Nov 20, 2017, at 4:53 PM, Michael Tüxen <notifications@github.com> wrote:
> 
> I don't see how the initialization can be done at compile time in contrast to run time.
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

--------
Randall Stewart
rrs@netflix.com
803-317-4952

tuexen:
I agree completely with Randy... The suggested change was only for userland examples. But I fail to see the advantage...

lgrahl:
The only real advantage is that it **could** result in more optimised assembly instructions.

Downsides are what you and Randall said.

PR #185, opened 2017-11-19
SCTP_DEBUG_ALL

JlnWntr:
Does it make sense to 
```
#ifdef SCTP_DEBUG
	usrsctp_sysctl_set_sctp_debug_on(SCTP_DEBUG_NONE);
#endif
```

?

### DIFF
--- a/programs/client.c
+++ b/programs/client.c
@@ -319,7 +319,7 @@ main(int argc, char *argv[])
 		usrsctp_init(9899, NULL, debug_printf);
 	}
 #ifdef SCTP_DEBUG
-	usrsctp_sysctl_set_sctp_debug_on(SCTP_DEBUG_NONE);
+	usrsctp_sysctl_set_sctp_debug_on(SCTP_DEBUG_ALL);
 #endif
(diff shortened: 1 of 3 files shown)


lgrahl:
Interesting find. Turning off debugging when `SCTP_DEBUG` is set doesn't make any sense to me.

tuexen:
The default value is SCTP_DEBUG_NONE. So there is no functional change. On the one hand side it is good to be consistent (so adding this would be good), on the other hand the call is completely optional... Just as a side note: The value is a bitmask... SCTP_DEBUG_NONE doesn't enable any output, SCTP_DEBUG_ALL enables all. But you can set it to other values as well.

lgrahl:
> The default value is SCTP_DEBUG_NONE. So there is no functional change.

That's why it puzzled me.

JlnWntr:
Mr. Tuexen explained to me today, that it makes sense to _enable_ SCTP_DEBUG (during compilation) but (at the same time) be able to set the explicit "debug-level". SCTP_DEBUG_NONE (with no  debug-output) is one level/option.

PR #189, opened 2017-12-11
Upcall API and some examples (2 clients and 4 servers)

ruengeler:


lgrahl:
@weinrank Leaving a note here that we wanted to take a look at this during IETF 101.

lgrahl:
Also linking https://github.com/NEAT-project/usrsctp-neat/issues/4

PR #202, opened 2018-02-15
Fix data race around timer_thread_should_exit.

### LINKED ISSUE #201: Data race around "timer_thread_should_exit"
ThreadSanitizer caught a data race between `sctp_finish` setting `timer_thread_should_exit` and `user_sctp_timer_iterate` reading it. It's effectively the same as this example ["Notification" data race](https://github.com/google/sanitizers/wiki/ThreadSanitizerPopularDataRaces#notification). 

Apparently it's legal for the compiler to rewrite the timer loop to be:

```
	if (!SCTP_BASE_VAR(timer_thread_should_exit)) {
		for (;;) {
			Sleep(TIMEOUT_INTERVAL);
			sctp_handle_tick(MSEC_TO_TICKS(TIMEOUT_INTERVAL));
		}
	}
```

But I've never seen this happen in practice.

Still, it would be nice to avoid the ThreadSanitizer warning. Is there any precedent for adding memory barriers to fix races like this? I notice user_atomic.h is only used in a couple places.

taylor-b:
Fixes #201.

### DIFF
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -46,6 +46,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <errno.h>
+#include <user_atomic.h>
 #include <netinet/sctp_sysctl.h>
 #include <netinet/sctp_pcb.h>
 #else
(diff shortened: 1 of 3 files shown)


taylor-b:
@tuexen, can you take a look?

PR #210, opened 2018-03-27
Fix increment in sctp_connectx_helper_add

taylor-b:
Increment should use sizeof(struct sockaddr_conn) instead of
sizeof(struct sockaddr_in6).


taylor-b:
@tuexen Can you take a look? This and https://github.com/sctplab/usrsctp/pull/209 were found when trying to build without IPv6 support.

tuexen:
@taylor-b Sure. Both look good and are merged. Thanks a lot!

PR #218, opened 2018-04-27
sctp_finish: wait thread join regardless socket status

justinjoy:
It has no chance to call tread join because of checking
sockets validity, but the sockets are already invalidated
by `recv_thread_destroy`

### DIFF
--- a/usrsctplib/netinet/sctp_pcb.h
+++ b/usrsctplib/netinet/sctp_pcb.h
@@ -286,6 +286,13 @@ struct sctp_epinfo {
 #endif
 };
 
+enum sctp_base_thread {
+        SCTP_BASE_THREAD_RECV_ROUTE = 0,
+        SCTP_BASE_THREAD_RECV_RAW,
+        SCTP_BASE_THREAD_RECV_UDP,
+        SCTP_BASE_THREAD_RECV_RAW6,
+        SCTP_BASE_THREAD_RECV_UDP6,
+};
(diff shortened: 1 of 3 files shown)


tuexen:
But what  happens if you did not start the thread? You can't join it in that case. In the case you did not start the thread, the fd are -1.

tuexen:
Rethinking about https://github.com/sctplab/usrsctp/commit/dd07478134d47d136858d6945083697701ef9eb6. How do you deal with the case not all sockets could be opened (like raw sockets and you don't have the privilege to do so). Then the fd is -1 and the corresponding receive thread wouldn't be started...

justinjoy:
good point. I didn't notice the case that threads couldn't be started.
In that case, I think `sctp_base_info` should have another member to trace if each thread is running.

usrsctplib/netinet/sctp_pcb.h
```diff
@@ -345,6 +352,7 @@ struct sctp_base_info {
 	void (*debug_printf)(const char *format, ...);
 	int crc32c_offloaded;
 #endif
+        int userland_thread_stats;
```
justinjoy: I was thinking, if I need to add something to here for tracing thread status, it would be better to be a single variable than multiple variables(per thread). However, I am not sure if this approach is right.

justinjoy:
Proposed different approach to  #229

PR #223, opened 2018-05-14
Make MTU 4-byte aligned

ruengeler:


tuexen:
Fixed in a different way in https://github.com/sctplab/usrsctp/commit/46a0fd5dcf888f5bc52b42959b61faab43be8c5a.

PR #229, opened 2018-05-22
join user thread by checking flag

justinjoy:
Invalidating socket was introduced by dd07478 to prevent
using closed socket in user thread, but it was reverted
by ec8cf8e because the socket descriptor is also used for
checking if user thread is created.

However, it sometimes happens that a new thread is created while
the existing thread is closing. Then, in user thread function, the
socket descriptor is regarded as valid so the loop in the function
has no chance to exit. To avoid replacing the socket descriptor
by repeating, the descriptor should be invalidated immediately.

Signed-off-by: Justin Kim <justin.kim@collabora.com>

### DIFF
--- a/usrsctplib/netinet/sctp_pcb.h
+++ b/usrsctplib/netinet/sctp_pcb.h
@@ -317,6 +317,7 @@ struct sctp_base_info {
 #if defined(INET) || defined(INET6)
 	int userspace_route;
 	userland_thread_t recvthreadroute;
+	int recvthreadroute_should_exit;
 #endif
 #endif
 #ifdef INET
@@ -329,6 +330,9 @@ struct sctp_base_info {
 #endif
 	userland_thread_t recvthreadraw;
 	userland_thread_t recvthreadudp;
+	int recvthreadraw_should_exit;
+	int recvthreadudp_should_exit;
+
 #endif
 #ifdef INET6
 #if defined(__Userspace_os_Windows)
@@ -340,6 +344,8 @@ struct sctp_base_info {
 #endif
 	userland_thread_t recvthreadraw6;
 	userland_thread_t recvthreadudp6;
+	int recvthreadraw6_should_exit;
+	int recvthreadudp6_should_exit;
 #endif
 	int (*conn_output)(void *addr, void *buffer, size_t length, uint8_t tos, uint8_t set_df);
 	void (*debug_printf)(const char *format, ...);
--- a/usrsctplib/netinet/sctp_usrreq.c
+++ b/usrsctplib/netinet/sctp_usrreq.c
@@ -136,15 +136,20 @@ sctp_init(void)
 #if !defined(__Userspace_os_Windows)
 #if defined(INET) || defined(INET6)
 	SCTP_BASE_VAR(userspace_route) = -1;
+	SCTP_BASE_VAR(recvthreadroute_should_exit) = 0;
 #endif
 #endif
 #ifdef INET
 	SCTP_BASE_VAR(userspace_rawsctp) = -1;
 	SCTP_BASE_VAR(userspace_udpsctp) = -1;
+	SCTP_BASE_VAR(recvthreadraw_should_exit) = 0;
+	SCTP_BASE_VAR(recvthreadudp_should_exit) = 0;
 #endif
 #ifdef INET6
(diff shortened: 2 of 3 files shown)


tuexen:
Can you provide an example where you create a new thread while the old one is still there? I think `sctp_finish()` returns when all threads have terminated. So calling `sctp_init()` after it should be fine.

justinjoy:
It happens when gstreamer pipeline is created and destroyed. In gstreamer plugin,  `usrsctp_init()`[0] and `usrsctp_finish()`[1] are called as per the instance lifecycle, but we can create multiple instances of gstreamer plugin so I think `usrsctp_init()` should be called once in a process.

[0] https://github.com/EricssonResearch/openwebrtc-gst-plugins/blob/master/ext/sctp/sctpassociation.c#L161
[1] https://github.com/EricssonResearch/openwebrtc-gst-plugins/blob/master/ext/sctp/sctpassociation.c#L196

tuexen:
You need to call `sctp_init()` once to initialise the stack. If you want it teardown the stack, you need to do something like
```
        while (usrsctp_finish() != 0) {
#ifdef _WIN32
                Sleep(1000);
#else
                sleep(1);
#endif
        }
```
Once `usrsctp_finish()` returned with 0, the stack is torn down. Only then, you should call `sctp_init()` again, if necessary. However, the basic idea is, that you call `sctp_init()` once when the program starts, `usrsctp_finish()` once when the program terminates. In between endpoints and associations can come and go...

gkiagia:
> ```
>       while (usrsctp_finish() != 0) {
>#ifdef _WIN32
>               Sleep(1000);
>#else
>                sleep(1);
>#endif
>        }
>```

Ugh, is there a good reason why usrsctp_finish() is not deterministic? Adding a random sleep() call there looks like a terrible idea.

>However, the basic idea is, that you call sctp_init() once when the program starts, usrsctp_finish() once when the program terminates. In between endpoints and associations can come and go...

This is not realistic when you are not writing the "program" itself, but a library that builds on top of this library, like we do in GStreamer. The GStreamer sctp plugin is just a plugin... The actual program can load and unload it at will. The plugin cannot possibly have control over static program initialization / deinitialization and even if it could have, it is bad design to try and outsmart the program. The program should be in control over what is loaded and what is not. If it wishes to unload something, it should be possible for it to do so without leaving anything in memory.

tuexen:
`usrsctp_finish()` can only do its job if there are no associations anymore and no endpoints.
So you could call `usrsctp_close()` on all your sockets but it might take a while until the associations have actually been teared down. That is what you have to compensate for. If you as an upper layer of SCTP make sure that all associations and end-points are gone before you call `usrsctp_finish()`, you don't need to wait.
There is currently no way to free all resources at arbitrary times... That would requires to terminate all associations (easy to do), remove all sockets (harder to do since the application controls these). I guess the best way would be that if the application wants to terminate the stack, it terminates all associations in a non-graceful way, closes all sockets and call `usrsctp_finish()`.

lgrahl:
1. @justinjoy Is this a WebRTC data channel implementation?
2. @tuexen Will `usrsctp_finish()` return successfully when all associations have been teared down non-gracefully (using ABORT)?

If the answer to both questions is yes, you can always [use an ABORT chunk](https://github.com/rawrtc/rawrtc/blob/master/src/librawrtc/sctp_transport.c#L2172:L2179) to tear down your associations and in fact [the W3C WebRTC API now requires you to do so](https://rawgit.com/w3c/webrtc-pc/master/webrtc.html#dom-rtcpeerconnection-close) (see step 7).

tuexen:
@lgrahl `usrsctp_finish()` will not trigger the teardown. It will just check if there are no end points and associations and if that is true, teardown the stack. That is why you have to retry until you are successful.

lgrahl:
It doesn't trigger the teardown but if all associations have been teared down prior to that with `usrsctp_close()` using an ABORT, will `usrsctp_finish()` return successfully?

tuexen:
@lgrahl That is what I would expect the stack to do.

justinjoy:
Another reason is that I'd like to avoid reusing a socket descriptor accidentally in a loop, but I hesitated to send it because it can be done by adding an atomic comparator(a361c2d) every turn in a loop. Please, think again about my PR. Thanks.

PR #242, opened 2018-06-15
Fix a build fails with gcc 4.5

s-hashimoto00:
`#pragma GCC diagnostic push` and `#pragma GCC diagnostic pop` were added in [gcc 4.6](https://gcc.gnu.org/gcc-4.6/changes.html). 

see also:
https://stackoverflow.com/questions/16555585/why-pragma-gcc-diagnostic-push-pop-warning-in-gcc-c
https://stackoverflow.com/questions/28681305/pragma-ignored-in-g-and-clang

### DIFF
--- a/usrsctplib/user_socket.c
+++ b/usrsctplib/user_socket.c
@@ -3519,7 +3519,7 @@ int usrsctp_sysctl_set_ ## __field(uint32_t value)   \
 	}                                            \
 }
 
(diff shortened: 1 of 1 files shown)


weinrank:
@tuexen Windows is fine with it!
Tested on Windows 10 with MSVC 2017

tuexen:
@Weinrank Thanks for testing.

PR #243, opened 2018-06-22
Switching from localtime to localtime_r on non-Windows platforms.

taylor-b:
localtime isn't thread-safe, since it returns a pointer to a
statically-allocated struct, whereas localtime_r takes the address of a
user-allocated struct.

This is only used in usrsctp_dumppacket.


tuexen:
Thanks for the patch.

PR #257, opened 2018-09-11
Minor build system fixes

somtin:
Hi,

These are some minor changes to the build system so that:

- Building the example programs can be disabled
- autoreconf can be used to initialise the build system instead of using the ./bootstrap script as autoreconf correctly uses the environment supplied by Yocto (and boostrap.sh does not)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,7 +28,11 @@
 # SUCH DAMAGE.
 #
 
-SUBDIRS = usrsctplib programs
+if COND_PROGRAMS
+MAYBEPROGRAMS = programs
+endif
+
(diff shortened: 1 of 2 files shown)


tuexen:
@weinrank Can we have a similar capability for the `cmake` based toolchain?

tuexen:
@somtin Thanks!

weinrank:
@tuexen It's already there 😉 
https://github.com/sctplab/usrsctp/blob/master/CMakeLists.txt#L74

tuexen:
@weinrank Great, thanks!

PR #273, opened 2018-10-25
Fixed notify happening before association is added to accept queue

somtin:
When the server socket is operating in non-blocking mode, we wait for
the COMM_UP notification, and then call accept. However, the
association is not added to the accept queue until after the COMM_UP
notification is triggered. This causes the application to have to wait an
indeterminate amount of time and can cause race conditions when the
application calls accept before usrSCTP adds the association to the
accept queue.

### DIFF
--- a/usrsctplib/netinet/sctp_input.c
+++ b/usrsctplib/netinet/sctp_input.c
@@ -3096,11 +3096,6 @@ sctp_handle_cookie_echo(struct mbuf *m, int iphlen, int offset,
 			SCTP_INP_DECR_REF(inp);
 			/* Switch over to the new guy */
 			*inp_p = inp;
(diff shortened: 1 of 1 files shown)


tuexen:
Could you provide a minimal test program which doesn't work before you change and does work after it?

somtin:
I'm working on something else currently, but if I can find some time I'll do that. Not sure how long that'll be though.

AlmightyChin:
Hi @tuexen ... Apologies for the delay. I've created an example using the ekr example program. unzip and replace ekr_server.c, and run the server and client as normal. With the sleep, it spams the console with "usrsctp_accept: Resource temporarily unavailable" until the function returns and adds the association to the accept queue. Without the sleep, it depends on when the OS schedules the main thread to run - sometimes it fails until the second try, sometimes it works.

[ekr_server.zip](https://github.com/sctplab/usrsctp/files/3937416/ekr_server.zip)

PR #298, opened 2019-02-04
Fix missed usrsctp_sysctl_set_sctp_multiple_asconfs

nxrighthere:
`usrsctp_sysctl_set_sctp_multiple_asconfs()` function definition is missed which makes it impossible to use from outside.


tuexen:
This sysctl is not exported since it adds functionality not allowed by the RFC. So this is intentionally...

nxrighthere:
I see, thanks for clarifying this.

PR #300, opened 2019-02-17
Add Meson build system support

lgrahl:
Adds [Meson build system](https://mesonbuild.com/) support. Tested with:

* Linux x64 (Arch, Ubuntu Trusty/Xenial/Bionic)
* Linux ARMv6 and ARMv7 (cross-compiled)
* Mac OS X Xcode 9.4 and Xcode 10.1
* Android API level 28 (cross-compiled)
* Windows x86 and x64 (cross-compiled via MinGW)
* Windows x86 and x64


lgrahl:
Updated the previous post with my build results.

tuexen:
@lgrahl MinGW support has been added by @weinrank . Can you check if your Meson build works on MinGW. It might be necessary to add something like https://github.com/sctplab/usrsctp/pull/321/commits/111e0caa5911833ef223df0ec92546582c8d83c7 to work around a gcc issue on Windows.

Let me know when you think this is ready to merge.

lgrahl:
Will rebase and test again. :+1:

tuexen:
> Will rebase and test again. 👍

Great, thanks.

lgrahl:
Updated. Results are [here (122-136)](https://circleci.com/gh/rawrtc/usrsctp/tree/meson-ci), [here](https://ci.appveyor.com/project/lgrahl/usrsctp/builds/25779471) and [here](https://travis-ci.org/rawrtc/usrsctp/builds/554803559).

This should be good to merge now. A couple of notes:

* Android builds only work with `meson configure -Dsctp_build_programs=false` since it does not support `pthread_cancel`. And there are a few warnings but nothing that raises any eyebrows.
* When building with MinGW, `inet_pton` was not available. I have resolved this by defining `_WIN32_WINNT=0x601`. You may want to copy this to the CMake and autotools variants.
* Building with MSVC on Windows only works with `meson configure -Dsctp_build_programs=false` or `meson configure --default-library=static` since usrsctp currently does not explicitly export symbols.

You can also take (and extend) my CI files of the [meson-ci](https://github.com/rawrtc/usrsctp/commit/f04f4eae42bfa591fc1e60866d5876c890f3397d) branch if you want to. I'd imagine continuous integration would greatly help in maintaining support for all platforms.

tuexen:
> Updated. Results are [here (122-136)](https://circleci.com/gh/rawrtc/usrsctp/tree/meson-ci), [here](https://ci.appveyor.com/project/lgrahl/usrsctp/builds/25779471) and [here](https://travis-ci.org/rawrtc/usrsctp/builds/554803559).
> 
> This should be good to merge now. A couple of notes:

Merged. Thanks for providing it.

> 
> * Android builds only work with `meson configure -Dsctp_build_programs=false` since it does not support `pthread_cancel`. And there are a few warnings but nothing that raises any eyebrows.

OK. The test programs are not really intended to be used on Android. So this is acceptable.

> * When building with MinGW, `inet_pton` was not available. I have resolved this by defining `_WIN32_WINNT=0x601`. You may want to copy this to the CMake and autotools variants.

@weinrank Do you observe the same problem? Why does defining `_WIN32_WINNT=0x601` makes a difference? According to [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-inet_pton), the function is supported on Windows 8.1 and Vista. That doesn't fit to using 0x601, which is according to [MSDN](https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=vs-2019) `_WIN32_WINNT_WIN7`.

> * Building with MSVC on Windows only works with `meson configure -Dsctp_build_programs=false` or `meson configure --default-library=static` since usrsctp currently does not explicitly export symbols.
> 

@weinrank Is this need on Windows? Should  we do that?

> You can also take (and extend) my CI files of the [meson-ci](https://github.com/rawrtc/usrsctp/commit/f04f4eae42bfa591fc1e60866d5876c890f3397d) branch if you want to. I'd imagine continuous integration would greatly help in maintaining support for all platforms.

@weinrank Any interest?

PR #319, opened 2019-06-12
Fix crash with HTCP

nxrighthere:
Fixes https://github.com/sctplab/usrsctp/issues/395

### DIFF
--- a/usrsctplib/netinet/sctp_cc_functions.c
+++ b/usrsctplib/netinet/sctp_cc_functions.c
@@ -1962,7 +1962,7 @@ htcp_cong_time(struct htcp *ca)
 static inline uint32_t
 htcp_ccount(struct htcp *ca)
 {
-	return (htcp_cong_time(ca)/ca->minRTT);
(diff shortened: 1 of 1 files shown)


lgrahl:
Ran into the same issue and applied the patch successfully.

ghazel:
Same here.

saurik:
@tuexen Yo! Just in case, I thought I'd directly ping you; there seems to be four people (I'm working with @ghazel, and so don't count as another person myself, but there are three in this thread and one in the referenced issue) who have experienced the same bug and are saying this patch works... what is the best way to help get this change committed?

tuexen:
The reason I was hesitating to commit this, is that the CC is not tested very well and the corresponding CC for TCP implemented in [cc_htcp.c](https://svnweb.freebsd.org/base/head/sys/netinet/cc/cc_htcp.c) seems to be different. It is based on an individual submission ([draft-leith-tcp-htcp-06.txt](https://tools.ietf.org/html/draft-leith-tcp-htcp-06)) and I'm not sure why it wasn't adopted in a WG and further developed to an RFC.

However, this pull requested avoids a division by zero bug, so it is better what was there before.

PR #320, opened 2019-06-14
Fix cross-build linux->mingw

orgads:


tuexen:
@weinrank Can you check this doesn't break a native Windows build?

orgads:
ping?

This cannot break Windows build, as Windows is case insensitive. I built the project on Windows, and it's fine.

weinrank:
I've quickly checked the PR (configure & compile) and it looks good to me.

tuexen:
> ping?
> 
> This cannot break Windows build, as Windows is case insensitive. I built the project on Windows, and it's fine.

Sorry. I was confused by the MSDN documentation, which sometimes use the first letter as a capital letter, but their examples do not... Thanks for the fix.

Issue #276, opened 2018-11-09
MinGW support


nxrighthere:
I see a bunch of pull requests related to MinGW support, but they are quite outdated. Any plans to support MinGW out of the box, please, to avoid headaches with digging the code?

tuexen:
I picked from the patches what I found generic improvements. I found contradicting change requests in different pull request and, since we are not windows experts, have no MinWG environment. If I remember it correctly, I think when we were looking around to set one up, we found that there are multiple ways of doing so and it wasn't clear which one to use.

Is there one MinGW environment we should use? Getting something in I can't test is something I'm hesitating to do.

CoolmanCZ:
Hi,
I think that [Appveyor](https://www.appveyor.com) can be used to test windows version (at least for the building). It has an integration with GitHub.

nxrighthere:
> Is there one MinGW environment we should use? 

Most people are using [this](https://sourceforge.net/projects/mingw-w64/) distribution. 

It would be great if you can separate the code in the traditional way using directives like `#if defined(_WIN32) && defined(__GNUC__)` instead of `__MINGW32__` for better compatibility. Right now GCC throws a lot of errors. Some of them are quite easy to fix, but some are not, unfortunately.

> I think that Appveyor can be used to test windows version (at least for the building).

Thanks for the suggestion.

CoolmanCZ:
If you are looking for mingw version of the usrsctp take a look at [my MinGW fork](https://github.com/CoolmanCZ/usrsctp). I've been trying to keep the source code updated with latest version of the usrsctp.

nxrighthere:
@CoolmanCZ Thanks! I've built debug version without any problems, but release throws some errors, unfortunately.

CoolmanCZ:
Hi, can you send me the command to build the release version?

nxrighthere:
You can set it right in the CMake:

![release](https://user-images.githubusercontent.com/8279330/48259771-e372fc80-e43a-11e8-9ab6-6fbf68131c5f.PNG)

CoolmanCZ:
OK, try fixed Mingw branch.

nxrighthere:
That was fast, thank you, it works. 😺

weinrank:
We already have a Windows buildbot with MinGW support, but we haven't used it yet.

I'll have a closer look at your branch during the next days and we'll check how to integrate it. 👍

nxrighthere:
@weinrank Any progress on this? 🎗

[...]

PR #321, opened 2019-06-28
MinGW Support (32bit & 64bit)

weinrank:
This PR adds support for MinGW (32bit & 64bit).
I'm not an MinGW expert so I've just done whatever seemed to be the right way.

I've taken some code from PR #150 and extended/modified it.
Tested in all combinations on Windows 10 (1903)
* MinGW32 & MinGW64
* Clang (i386 / x64)
* GCC (i386 / x64)

If this PR gets accepted, I'll add two additional Buildslaves to the CI system.

References:
#300 - Meson build system will probably need some extensions for MinGW support
#276 - Should be fixed now, I've used `__MINGW32__` in favour of the suggested `__GNUC__` 
#150 - Original PR which should be obsoleted by this PR


programs/discard_server.c
```diff
@@ -105,7 +105,7 @@ receive_cb(struct socket *sock, union sctp_sockstore addr, void *data,
 			       rcv.rcv_sid,
 			       rcv.rcv_ssn,
 			       rcv.rcv_tsn,
-			       ntohl(rcv.rcv_ppid),
```
tuexen: Why are you casting the result? `ntohl()` and alike return an unsigned integer: [Posix](https://pubs.opengroup.org/onlinepubs/009695399/functions/htonl.html). This comment applies also to other similar changes.

weinrank: To fix a compiler warning for MinGW with GCC, clang is fine with the code without casting it.
`htonl()` on Windows returns an `u_long` [docs here](https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-htonl)

`u_long` gets translated to `unsigned long`

Since it doesn't harm the any other platform, I didn't write another workaround.

```
[ 49%] Building C object programs/CMakeFiles/discard_server_upcall.dir/discard_server_upcall.c.obj
C:/Documents and Settings/weinrank/Documents/GitHub/usrsctp/programs/discard_server_upcall.c: In function 'handle_upcall':
C:/Documents and Settings/weinrank/Documents/GitHub/usrsctp/programs/discard_server_upcall.c:124:93: error: format '%u' expects argument of type 'unsigned int', but argument 8 has type 'u_long' {aka 'long unsigned int'} [-Werror=format=]
  124 |     printf("Msg of length %d received from %s:%u on stream %d with SSN %u and TSN %u, PPID %u, context %u.\n",
      |                                                                                            ~^
      |                                                                                             |
      |                                                                                             unsigned int
      |                                                                                            %lu
......
  131 |            ntohl(rn.recvv_rcvinfo.rcv_ppid),
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |            |
      |            u_long {aka long unsigned int}
cc1.exe: all warnings being treated as errors
make[2]: *** [programs/CMakeFiles/discard_server_upcall.dir/build.make:63: programs/CMakeFiles/discard_server_upcall.dir/discard_server_upcall.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:301: programs/CMakeFiles/discard_server_upcall.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

programs/rtcweb.c
```diff
@@ -120,10 +120,10 @@ struct peer_connection {
 #define SCTP_PACKED
 #endif
 
-#if defined(_WIN32)
+#if defined(_WIN32) && !defined(__MINGW32__)
 #pragma warning( push )
 #pragma warning( disable : 4200 )
-#endif //defined(_WIN32)
```
tuexen: While you are there: Can you change it to a C-style comment?

usrsctplib/user_socket.c
```diff
@@ -3138,14 +3138,14 @@ sctp_userspace_ip_output(int *result, struct mbuf *o_pak,
 	if ((!use_udp_tunneling) && (SCTP_BASE_VAR(userspace_rawsctp) != -1)) {
 		if (WSASendTo(SCTP_BASE_VAR(userspace_rawsctp), (LPWSABUF) send_iovec, iovcnt, &win_sent_len, win_msg_hdr.dwFlags, win_msg_hdr.name, (int) win_msg_hdr.namelen, NULL, NULL) != 0) {
 			*result = WSAGetLastError();
-		} else if (win_sent_len != send_len) {
```
tuexen: Please no space between `(type)` and `variable` when casting... Applies multiple times.

programs/discard_server.c
tuexen: I don't see a problem here. Windows uses `unsigned long`, which is the same as `unsigned int`, since Windows is a LLP64 system. Both are 32-bit unsigned integers. System other than Windows often use ILP32, which means the `unsigned int` is a 32-bit integer and `unsigned long` is a 64-bit entity.

So I guess this is a bug in gcc. Can we leave the code untouched, since it is correct, and just turn off the warning when using gcc on Windows?

programs/rtcweb.c
weinrank: Done!

usrsctplib/user_socket.c
weinrank: done!

weinrank:
Comments are addressed.

PR #328, opened 2019-07-24
Fix MSVC Meson build

xhaakon:
From the output of "dumpbin /linkermember" run on the static version of usrsctp generate .def file with function names to export from the DLL.

### DIFF
--- a/gen-def.py
+++ b/gen-def.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python3
+#
+# gen-def.py usrsctp.lib
+import re
+import sys
+import subprocess
+from shutil import which
+
+try:
+    lib_file = sys.argv[1]
+except:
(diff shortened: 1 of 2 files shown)


tuexen:
@lgrahl : OK to merge?

lgrahl:
I'll run this with my CI configuration first and will report back.

xhaakon:
> Static builds break with this patch. I wrote a patch to fix that: rawrtc@d1d0561

Thanks! I squashed the fix with my commit (and mentioned you as the de facto co-author of the change).

> Windows x86 builds are broken now

Apparently function names in x86 static lib are prefixed with an underscore. Fixed the regular expression in get-def.py accordingly.

> The commit message should describe

I extended the commit description a bit.

tuexen:
@lgrahl Thanks for testing.

PR #329, opened 2019-07-29
MinGW Compile fix

weinrank:
(Hopefully) solves #276


usrsctplib/netinet/sctp_pcb.c
```diff
@@ -6852,7 +6852,9 @@ sctp_pcb_init()
 #endif
 #if defined(__Userspace__)
 	mbuf_initialize(NULL);
+#if !defined(__MINGW32__)
```
tuexen: Please don't insert an `ifdef` here. Let us call `atomic_init()` in any case for the userland stack.

Please replace in `user_atomic.h`
```
#if defined(__Userspace_os_Windows)
#if !defined(__MINGW32__)
static void atomic_init(void) {} /* empty when we are not using atomic_mtx */
#endif
#else
static inline void atomic_init(void) {} /* empty when we are not using atomic_mtx */
#endif
```
by
```
#if defined(__Userspace_os_Windows)
static void atomic_init(void) {} /* empty when we are not using atomic_mtx */
#else
static inline void atomic_init(void) {} /* empty when we are not using atomic_mtx */
#endif
```

weinrank: Done!

tuexen: Thanks a lot!

PR #334, opened 2019-08-01
Improved ekr_loop logging

weinrank:


programs/ekr_loop.c
```diff
@@ -487,24 +490,29 @@ main(int argc, char *argv[])
 	sndinfo.snd_ppid = htonl(DISCARD_PPID);
 	sndinfo.snd_context = 0;
 	sndinfo.snd_assoc_id = 0;
+
 	for (i = 1; i < NUMBER_OF_STEPS; i++) {
```
tuexen: I think this should be `i = 0`.

weinrank: fixed!

PR #356, opened 2019-08-23
Fuzzer integration

weinrank:
This PR integrates a standalone fuzzing environment as well as a preparation for Google's OSS-FUZZ project.


benjwright:
Great work! Thanks for this change Felix.

Issue #352, opened 2019-08-21
CPU DoS (infinite loop) in sctp_send_cookie_echo


markwo:
While working on a fuzzer for usrsctp, I hit a timeout due to an infinite loop in the target/receiver endpoint. The PCAP is attached - output and stack trace of the timeout are below. From a look at the code, this is due to the code in sctp_send_cookie_echo (sctp_output.c) not handling parameters with a zero length:

```
        for (;;) {
                phdr = sctp_get_next_param(m, at, &param, sizeof(param));
                if (phdr == NULL) {
                        return (-3);
                }
                ptype = ntohs(phdr->param_type);
                plen = ntohs(phdr->param_length);
                if (ptype == SCTP_STATE_COOKIE) {
                ...
                }
                at += SCTP_SIZE32(plen);
```

I'm cleaning up the fuzzer currently - can share the code if you need it to repro. (Would also like to eventually get this fuzzer added to the repo as an OSS-Fuzz target).

```
ALARM: working on the last Unit for 5 seconds
       and the timeout value is 5 (use -timeout=N to change)
==102223== ERROR: libFuzzer: timeout after 5 seconds
    #0 0x55a619edd351 in __sanitizer_print_stack_trace third_party/llvm/llvm/projects/compiler-rt/lib/asan/asan_stack.cpp:86:3
    #1 0x55a61a0d8bc8 in fuzzer::PrintStackTrace() third_party/llvm/llvm/projects/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:205:5
    #2 0x55a61a0bcae6 in fuzzer::Fuzzer::AlarmCallback() third_party/llvm/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:300:5
    #3 0x7f5b5a73599f  (/usr/grte/v4/lib64/libpthread.so.0+0xf99f)
    #4 0x55a619fc3541 in sctp_send_cookie_echo third_party/usrsctp/usrsctplib/netinet/sctp_output.c:9495:10
    #5 0x55a619f5f5d2 in sctp_process_init_ack third_party/usrsctp/usrsctplib/netinet/sctp_input.c:580:11
    #6 0x55a619f5399c in sctp_handle_init_ack third_party/usrsctp/usrsctplib/netinet/sctp_input.c:1515:7
    #7 0x55a619f50c3f in sctp_process_control third_party/usrsctp/usrsctplib/netinet/sctp_input.c:5087:11
[...]

Issue #351, opened 2019-08-21
NULL deref crash in m_copydata


markwo:
While working on a fuzzer for usrsctp, I hit a SEGV accessing a NULL pointer. The PCAP is attached - output and stack trace of the crash are below. 

I'm cleaning up the fuzzer currently - can share the code if you need it to repro. (Would also like to eventually get this fuzzer added to the repo as an OSS-Fuzz target).

```
I 14:41:15.285924 0000 13 88 13 88 00 00 00 00 ab c0 54 a5 01 00 00 56 11 4f e7 17 00 02 00 00 00 04 00 04 4f d4 56 32 c0 00 00 04 80 08 00 09 c0 0f c1 80 82 00 00 00 80 02 00 24 1b 5e 9b 0f 4d 23 8a 48 8f 55 ab 62 e5 cf 31 6a 61 ad 64 37 db 19 0c 77 7f d3 1c 71 cc c0 2c 7d 80 04 00 06 00 01 00 00 80 03 00 06 80 c1 00 00 # SCTP_PACKET

O 14:41:15.286064 0000 13 88 13 88 00 00 00 00 00 00 00 00 01 00 00 14 01 00 00 00 00 00 20 00 00 08 00 08 00 00 00 01 # SCTP_PACKET

[...]

PR #358, opened 2019-08-29
Fixing issue 352

markwo:


tuexen:
I agree that this change would fix the issue.

https://github.com/sctplab/usrsctp/issues/351 and https://github.com/sctplab/usrsctp/issues/352 are incarnations of the same problem: a missing input validation. The root cause is the the input validation is done before and honours unknown parameters which request to stop processing parameters. This is done in `sctp_arethere_unrecognized_parameters()`. The lookup code in `sctp_send_cookie_echo()` does not honour the unknown parameters...
Besides the missing input validation, the code would also find cookie parameters it should not find.

Here is the fix I'm currently testing:

- Extend `sctp_arethere_unrecognized_parameters()` to indicate if a valid cookie was found. (this would fix both issues with input validation).
- Harden the scanning of parameters in `sctp_send_cookie_echo()` just in case it will be used in other contexts in the future. This requires protecting against zero length parameters like you suggest and partial parameters.

I'm planning to commit this later today after some more testing.

markwo:
Sounds good, thanks!

PR #360, opened 2019-08-30
Fix for issue#359. Guard stcb->sctp_ep when setting the 'connecting' …

dkisabaka:
Guard stcb->sctp_ep when setting the 'connecting' flag to prevent race conditions when two threads are binding and connecting to a socket.

### DIFF
--- a/usrsctplib/netinet/sctp_usrreq.c
+++ b/usrsctplib/netinet/sctp_usrreq.c
@@ -8179,11 +8179,13 @@ sctpconn_connect(struct socket *so, struct sockaddr *addr)
 		/* Gak! no memory */
 		goto out_now;
 	}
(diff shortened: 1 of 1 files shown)


usrsctplib/netinet/sctp_usrreq.c
```diff
@@ -8179,11 +8179,13 @@ sctpconn_connect(struct socket *so, struct sockaddr *addr)
 		/* Gak! no memory */
 		goto out_now;
 	}
-	if (stcb->sctp_ep->sctp_flags & SCTP_PCB_FLAGS_TCPTYPE) {
+	SCTP_INP_RLOCK(stcb->sctp_ep);
```
tuexen: You can't aquire an `inp` lock here since you own already the `stcb` lock. This would result in a LOR.

PR #364, opened 2019-09-12
Build fix for compilers without "-Wno-deprecated-declarations" option

dreibh:
Build fix: flags are already detected and set in ../CMakeLists.txt => build failed here if compiler does not support them.


tuexen:
I think this has already been fixed in the master branch. Please reopen, if the issue is still there.

PR #367, opened 2019-09-14
WIP: Problem: Cannot forward user data to the send callback.

### LINKED ISSUE #366: Problem: cannot forward user data to the send callback
Currently `usrsctp_socket` provides the ability to register a send and recv callback. The receive callback takes a user provided `void *ulp_info` which can be used to pass abitrary user. However, there is no equivalent mechanism for the send callback. Is there a specific reason why not?

My specific use case if to be able to application data that from a behind cffi where I can't simply put a global variable in the send callback.

Assuming this is a omission, what do you think of adding the following function?
```c
struct socket * usrsctp_socket2(
  int domain,
  int type,
  int protocol,
  int (*recv_callback)(
    struct socket *sock,
    union sctp_sockstore addr,
    void *data,
    size_t datalen,
    struct sctp_rcvinfo,
    int flags,
    void *recv_user_data
  ),
  int (*send_callback)(
    struct socket *sock,
    uint32_t sb_free,
    void *send_user_data,
  ),
  uint32_t sb_threshold,
  void *recv_user_data,
  void *send_user_data,
)
```

jean-airoldie:
~~Solution: Add `usrsctp_socket2` function which adds a send_callback2 which accepts the ulp_info. Add `usrsctp_set_assoc_sb_threshold` which allows configuration of the `sb_threshold` on a association basis, as opposed to a global sb_threshold.~~

~~The new signature would be:~~
```c
usrsctp_socket2(int domain, int type, int protocol,
               int (*receive_cb)(struct socket *sock, union sctp_sockstore addr, void *data, size_t datalen, struct sctp_rcvinfo, int flags, void *ulp_info),
               int (*send_cb)(struct socket *sock, uint32_t sb_free, sctp_assoc_t assoc_id, void *ulp_info),
               void *ulp_info);
```

~~This change should be backward compatible. Also, I couldn't think of a better name for the function.~~

This closes #366.

Edit: Outdated: See https://github.com/sctplab/usrsctp/pull/367#issuecomment-532907638

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,6 @@ callgrind.out*
 
 # cmake build
 /build
+
+# ctags
+tags
--- a/usrsctplib/netinet/sctp_indata.c
+++ b/usrsctplib/netinet/sctp_indata.c
@@ -4148,14 +4148,18 @@ sctp_express_handle_sack(struct sctp_tcb *stcb, uint32_t cumack,
 			sb_free_now = SCTP_SB_LIMIT_SND(stcb->sctp_socket) - (inqueue_bytes + stcb->asoc.sb_send_resv);
 
 			/* check if the amount free in the send socket buffer crossed the threshold */
-			if (inp->send_callback &&
+			if ((inp->send_callback || inp->send_callback2) &&
 			    (((inp->send_sb_threshold > 0) &&
 			      (sb_free_now >= inp->send_sb_threshold) &&
 			      (stcb->asoc.chunks_on_out_queue <= SCTP_BASE_SYSCTL(sctp_max_chunks_on_queue))) ||
 			     (inp->send_sb_threshold == 0))) {
 				atomic_add_int(&stcb->asoc.refcnt, 1);
 				SCTP_TCB_UNLOCK(stcb);
-				inp->send_callback(stcb->sctp_socket, sb_free_now);
+                if (inp->send_callback) {
+                    inp->send_callback(stcb->sctp_socket, sb_free_now);
+                } else {
+                    inp->send_callback2(stcb->sctp_socket, sb_free_now, stcb->sctp_ep->ulp_info);
+                }
 				SCTP_TCB_LOCK(stcb);
 				atomic_subtract_int(&stcb->asoc.refcnt, 1);
 			}
@@ -4872,12 +4876,16 @@ sctp_handle_sack(struct mbuf *m, int offset_seg, int offset_dup,
 			sb_free_now = SCTP_SB_LIMIT_SND(stcb->sctp_socket) - (inqueue_bytes + stcb->asoc.sb_send_resv);
 
 			/* check if the amount free in the send socket buffer crossed the threshold */
-			if (inp->send_callback &&
+			if ((inp->send_callback || inp->send_callback2) &&
 			   (((inp->send_sb_threshold > 0) && (sb_free_now >= inp->send_sb_threshold)) ||
 			    (inp->send_sb_threshold == 0))) {
 				atomic_add_int(&stcb->asoc.refcnt, 1);
 				SCTP_TCB_UNLOCK(stcb);
-				inp->send_callback(stcb->sctp_socket, sb_free_now);
+                if (inp->send_callback) {
+                    inp->send_callback(stcb->sctp_socket, sb_free_now);
+                } else {
+                    inp->send_callback2(stcb->sctp_socket, sb_free_now, stcb->sctp_ep->ulp_info);
+                }
 				SCTP_TCB_LOCK(stcb);
 				atomic_subtract_int(&stcb->asoc.refcnt, 1);
 			}
--- a/usrsctplib/netinet/sctp_input.c
+++ b/usrsctplib/netinet/sctp_input.c
@@ -3037,6 +3037,7 @@ sctp_handle_cookie_echo(struct mbuf *m, int iphlen, int offset,
 #if defined(__Userspace__)
 			inp->ulp_info = (*inp_p)->ulp_info;
 			inp->recv_callback = (*inp_p)->recv_callback;
+			inp->recv_callback2 = (*inp_p)->recv_callback2;
 			inp->send_callback = (*inp_p)->send_callback;
 			inp->send_callback2 = (*inp_p)->send_callback2;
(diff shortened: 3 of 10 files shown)


tuexen:
Why do you need a different pointer from the send callback than from the receive callback? I do understand that the upper layer wants a point and that it would make sense to have it for the send and receive callback. However, why do they need to be different?

jean-airoldie:
@tuexen At the time I wasn't aware that the send and receive callbacks shared the same lock so I wanted to avoid potential races. But you are right, there is no real reason to have two different pointers. I'll make the adjustments.

jean-airoldie:
I guess I'll also rename everything to `ulp_info`.

jean-airoldie:
Also couldn't think of a way to avoid code duplication between `usrsctp_socket` & `usrsctp_socket2` since from what I understand partial function application is not possible in c.

jean-airoldie:
So following by comments on #369, I added a per association send buffer threshold that can be configured at runtime. The rational is that using a global `sb_threshold` might cause a single stalling association to block all other associations. Indeed, the global sb_threshold does not give information to the send_callback about which association has free space in the send_buffer.

jean-airoldie:
For instance, if an application dedicates one thread for each association, the send_callback could be used to notify the appropriate thread that it is ready to write.

jean-airoldie:
I think the same concept should be applied to reads. The read_callback should be used to notify the appropriate association that it is ready to read.

@tuexen Do you think it would be possible to have a `recv` similar so `recvv` but that works on a per association basis?

jean-airoldie:
I guess at that point your better off using the `one-to-one` socket API.

jean-airoldie:
An alternative solution would be to use the `one-to-one` socket API and allow setting the `send_callback` & `recv_callback` for associations socket obtained via `accept`. This way multiple threads could each read and write their own association and they would each be independently notified via callbacks.

jean-airoldie:
I'll revert my change regarding the per association threshold.

jean-airoldie:
For a `one-to-one` sockets, the simplest API I can think of would be:
```c
typedef int (*recv_callback2_t)(struct socket *sock, uint32_t msg_size, void *ulp_info);
typedef int (*send_callback2_t)(struct socket *sock, uint32_t sb_free, void *ulp_info);

int
usrsctp_socket2(int domain, int type, int protocol, recv_callback2_t recv_callback,
    send_callback2_t send_callback, void *ulp_info);

struct socket *
usrsctp_accept2(struct socket *so, struct sockaddr * aname, socklen_t * anamelen,
    recv_callback2_t recv_callback, send_callback2_t send_callback);

int
usrsctp_set_sb_threshold(struct socket *, uint32_t threshold);
```

This would allow both `recvv` and `sendv` calls to be on association basis and not block each others. The `recv_callback` and `send_callback` would only be used for wakeups (as they should be because they acquire the `TCB_LOCK`).

For `one-to-many` socket, this would be significantly more complex, so I won't bother.

jean-airoldie:
But then accept is blocking, so it should take a callback instead of returning a socket.

jean-airoldie:
@tuexen I think I have a better understanding of the changes that would be required. Would appreciate your thoughts on the matter.

So rational behind the changes:

#### Add `send_callback2` and `recv_callback2`
  * Maintains backward compatibility, assuming it must be maintained.
  * Now both callbacks have access to `ulp_info`. Since `ulp_info` is behind a lock there should be no race issues.
  * The signature of `recv_callback2` was modified because its possible that a socket is ready to recv, but the user is not. The previous callback forced the user to recv the message, which might block in user code. Since a lock is acquired before either of the callback is called, they must not block in user code. My solution is to have callbacks simply acts as a dumb wakeup mechanism that can be used to delegate the read / write to another thread in user code.

The new signature is simply:
```c
/* The msg_size can be used by the user to judge whether they have enough
   space to recv the message without having to peek. */
void (*recv_callback2)(struct socket *sock, uint32_t msg_size, void *ulp_info),
void (*send_callback2)(struct socket *sock, uint32_t sb_free, void *ulp_info);
````
#### Add `set_send_callback2` & `set_recv_callback2`
  * This was mainly added to be able to configure a distinct callback from the original socket after calling `accept`. Ideally we would want an `accept2` function that accepts a `send_callback2` and `recv_callback2` but I wasn't able to figure that out. I don't think that it is an ideal solution because messages received before `recv_callback2` was set would not trigger the callback, but I'm not sure.

#### Unresolved questions
There is this one instance where a `recv_callback` because the association is down but no notification is passed to the user, instead NULL data and a zeroed `sctp_rcvinfo`. I'm not sure understand why.
https://github.com/sctplab/usrsctp/blob/8ca132191668c1dff35872315cf399cd7c3dba9b/usrsctplib/netinet/sctputil.c#L3863

jean-airoldie:
Additionally a non-blocking `accept2` coupled with a `listen2` method that takes an additionnal `accept_callback` that is run each time an association is queued for accept would be ideal. This way the user code would be waked up only when a connection can potentially be accepted.

tuexen:
> Additionally a non-blocking `accept2` coupled with a `listen2` method that takes an additionnal `accept_callback` that is run each time an association is queued for accept would be ideal. This way the user code would be waked up only when a connection can potentially be accepted.

Before adding more and more API's, I would prefer to make sure we can do what you want with the upcall API. I would imagine that you can just use the upcall API on the listening socket and if the socket is readable you can call accept.

I'm not sure if this is implemented yet, but if not, I would prefer to go that route.

The upcall API avoids a lot of problems (locking/memory management) for applications. So I would suggest to look into that direction.

tuexen:
Thinking more about the upcall API. It already has what you are trying to implement here: `usrsctp_set_upcall()` allows you to provide a pointer (like `ulp_info`) which you get provided on the callback, which generalises the `send()/recv()` callback.

Since I think the API to progress forward is the upcall API, I would pretty much invest effort to extend/debug that instead of adding more to the other APIs.

Can you have a look at the upcall API and report what is missing in it to be useable for you?

jean-airoldie:
@tuexen Cool, I'll look into.

jean-airoldie:
Seems like the `upcall` API is what I need. If there is anything missing I'll open a different PR. I plan to contribute in form of documentation once I'm done.

tuexen:
Documentation is really appreciated...

PR #369, opened 2019-09-14
Problem: Unable to modify sb_threshold at runtime

### LINKED ISSUE #368: Problem: Cannot update sb_threshold value
Currently the sb_threshold can only be set initially when a socket is created via `usrsctp_socket`. We should add a function to be able to update this value.

A possible use case would be to update the `sb_threshold` size from within the `send_callback` to avoid spurious wakeups.

I'll write a PR.

jean-airoldie:
Solution: Add usrsctp_set_sb_threshold function.

Closes #368.

### DIFF
--- a/usrsctplib/netinet/sctp_usrreq.c
+++ b/usrsctplib/netinet/sctp_usrreq.c
@@ -9073,4 +9073,19 @@ register_ulp_info (struct socket *so, void *ulp_info)
 	SCTP_INP_WUNLOCK(inp);
 	return (1);
 }
+
+int
+register_sb_threshold(struct socket *so, uint32_t sb_threshold)
+{
+	struct sctp_inpcb *inp;
+
+	inp = (struct sctp_inpcb *) so->so_pcb;
+	if (inp == NULL) {
+		return (0);
+	}
+	SCTP_INP_WLOCK(inp);
+	inp->send_sb_threshold = sb_threshold;
+	SCTP_INP_WUNLOCK(inp);
+	return (1);
+}
 #endif
--- a/usrsctplib/user_socket.c
+++ b/usrsctplib/user_socket.c
@@ -2585,6 +2585,12 @@ usrsctp_set_ulpinfo(struct socket *so, void *ulp_info)
 }
 
 int
+usrsctp_set_sb_threshold(struct socket *so, uint32_t sb_threshold)
+{
+    return (register_sb_threshold(so, sb_threshold));
+}
+
+int
 usrsctp_bindx(struct socket *so, struct sockaddr *addrs, int addrcnt, int flags)
 {
 	struct sctp_getaddresses *gaddrs;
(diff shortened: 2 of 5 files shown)


usrsctplib/netinet/sctp_usrreq.c
```diff
@@ -9073,4 +9073,19 @@ register_ulp_info (struct socket *so, void *ulp_info)
 	SCTP_INP_WUNLOCK(inp);
 	return (1);
 }
+int
+
+register_sb_threshold(struct socket *so, uint32_t sb_threshold)
+{
+	struct sctp_inpcb *inp;
+
```
tuexen: Can you add:
```
	if (so == NULL) {
		errno = EBADF;
		return (-1);
	}
```
as it is checked in most other functions?

usrsctplib/user_socket.c
```diff
@@ -2584,6 +2584,12 @@ usrsctp_set_ulpinfo(struct socket *so, void *ulp_info)
 	return (register_ulp_info(so, ulp_info));
 }
 
+int
+usrsctp_set_sb_threshold(struct socket *so, uint32_t sb_threshold)
+{
```
tuexen: Can you add
```
	if (so == NULL) {
		errno = EBADF;
		return (-1);
	}
```
as it is done for most other functions (and missing from `usrsctp_set_ulpinfo()`?

usrsctplib/usrsctp.h
```diff
@@ -1030,6 +1030,9 @@ usrsctp_deregister_address(void *);
 int
 usrsctp_set_ulpinfo(struct socket *, void *);
 
+int
+usrsctp_set_sb_threshold(struct socket *, uint32_t sb_threshold);
```
tuexen: Please remove `sb_threshold` for consistency.

jean-airoldie:
Now that I think about it, shouldn't `sb_threshold` be set on a association basis as there is a send_buffer for each association (from what i understand)? We would also need to add another `assoc_id` field to the `send_callback`.

jean-airoldie:
Otherwise a stalling association could potentially block all other associations.

jean-airoldie:
We would need to add the threshold field directly to the https://github.com/sctplab/usrsctp/blob/8ca132191668c1dff35872315cf399cd7c3dba9b/usrsctplib/netinet/sctp_pcb.h#L614
since it would be on an association basis.

jean-airoldie:
I'll close this issue since its too tightly linked with #367.

jean-airoldie:
Alright so I think the `one-to-many` API wouldn't work for my needs because I cannot chose from which association I recv a message using `recvv`. However, the `one-to-one` API could work.

Moreover, in the `one-to-one` API the `sb_threshold` acts as the threshold for that association, so we don't run into the problem I previously raised.

usrsctplib/usrsctp.h
jean-airoldie: Done.

tuexen: Really?

jean-airoldie: Right sorry, merged with the upstream.

jean-airoldie:
I'm no lounger using the callback API, in favor of the upcall API, so there is no point to this PR.

PR #373, opened 2019-09-17
Added missing SCTP_DATA_{LAST_FRAG,NOT_FRAG} flags

### LINKED ISSUE #372: Missing SCTP_DATA_{NOT_FRAG,LAST_FRAG} in header
These flags can be returned by a `SCTP_SEND_FAILED_EVENT` notification, in the `ssfe_info.sinfo_flags`, as defined in https://tools.ietf.org/html/rfc6458#section-6.1.11

jean-airoldie:
This closes #372.

### DIFF
--- a/usrsctplib/usrsctp.h
+++ b/usrsctplib/usrsctp.h
@@ -480,6 +480,8 @@ struct sctp_event_subscribe {
 
 
 /* Flags that go into the sinfo->sinfo_flags field */
+#define SCTP_DATA_LAST_FRAG   0x0001 /* tail part of the message could not be sent */
+#define SCTP_DATA_NOT_FRAG    0x0003 /* complete message could not be sent */
 #define SCTP_NOTIFICATION     0x0010 /* next message is a notification */
 #define SCTP_COMPLETE         0x0020 /* next message is complete */
 #define SCTP_EOF              0x0100 /* Start shutdown procedures */


usrsctplib/usrsctp.h
```diff
@@ -435,6 +435,8 @@ struct sctp_send_failed_event {
 /* flag that indicates state of data */
 #define SCTP_DATA_UNSENT  0x0001	/* inqueue never on wire */
 #define SCTP_DATA_SENT    0x0002	/* on wire at failure */
+#define SCTP_DATA_LAST_FRAG 0x01 /* tail part of the message could not me sent */
```
tuexen: I guess you mean "not be sent".

usrsctplib/usrsctp.h
```diff
@@ -435,6 +435,8 @@ struct sctp_send_failed_event {
...
+#define SCTP_DATA_NOT_FRAG 0x03 /* complete message could not be sent */
```
tuexen: I would prefer to have these definitions after
```
/* Flags that go into the sinfo->sinfo_flags field */
```
and before
```
#define SCTP_NOTIFICATION     0x0010 /* next message is a notification */
```
because both are used in `sinfo_flags`.

jean-airoldie: Sure thing.

jean-airoldie:
@tuexen Btw what are your thoughts on running clang-format against the repo?

tuexen:
> @tuexen Btw what are your thoughts on running clang-format against the repo?

We use the same sources, process them to take out various platform specific code, and then the resulting code is what is in FreeBSD. The code (more or less) is compliant with the coding style required for FreeBSD. For that reason, I would prefer not to run the code through an automated tool which might not be compliant with the FreeBSD coding style...

PR #377, opened 2019-09-20
Enable easier fuzzing of authenticated chunks

markwo:
This change is similar to the ones made by @weinrank to enable fuzzing.
Has no effect on normal builds.

### DIFF
--- a/usrsctplib/netinet/sctp_auth.c
+++ b/usrsctplib/netinet/sctp_auth.c
@@ -1730,6 +1730,7 @@ sctp_handle_auth(struct sctp_tcb *stcb, struct sctp_auth_chunk *auth,
 	(void)sctp_compute_hmac_m(hmac_id, stcb->asoc.authinfo.recv_key,
 	    m, offset, computed_digest);
 
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
 	/* compare the computed digest with the one in the AUTH chunk */
 	if (timingsafe_bcmp(digest, computed_digest, digestlen) != 0) {
 		SCTP_STAT_INCR(sctps_recvauthfailed);
@@ -1737,6 +1738,7 @@ sctp_handle_auth(struct sctp_tcb *stcb, struct sctp_auth_chunk *auth,
 			"SCTP Auth: HMAC digest check failed\n");
 		return (-1);
 	}
+#endif
 	return (0);
 }
 
--- a/usrsctplib/netinet/sctp_output.c
+++ b/usrsctplib/netinet/sctp_output.c
@@ -9537,6 +9537,9 @@ sctp_send_cookie_echo(struct mbuf *m,
 		}
 		ptype = ntohs(phdr->param_type);
 		plen = ntohs(phdr->param_length);
+		if (plen == 0) {
+			return (-3);
+		}
 		if (ptype == SCTP_STATE_COOKIE) {
 			int pad;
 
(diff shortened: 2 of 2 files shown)


usrsctplib/netinet/sctp_auth.c
```diff
@@ -1730,13 +1730,15 @@ sctp_handle_auth(struct sctp_tcb *stcb, struct sctp_auth_chunk *auth,
 	(void)sctp_compute_hmac_m(hmac_id, stcb->asoc.authinfo.recv_key,
 	    m, offset, computed_digest);
 
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
tuexen: I would like to see a change only in the userland stack. So Maybe something like
```
#if defined(__Userspace__)
#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
if (timingsafe_bcmp(digest, computed_digest, digestlen) != 0) {
#else
if (0) {
#endif
#else
if (timingsafe_bcmp(digest, computed_digest, digestlen) != 0) {
#endif
```

tuexen:
@markwo I added a similar change in https://github.com/sctplab/usrsctp/commit/d10edce494d79f33295914861a3be411f93ffe0e. Please let me know if this does not do what you want.

markwo:
The change looks good, thanks!

markwo:
I realized another change is helpful when fuzzing ASCONF chunks, also in sctp_asconf.c:
```
@@ -624,7 +624,11 @@ sctp_handle_asconf(struct mbuf *m, unsig
        }
        asoc = &stcb->asoc;
        serial_num = ntohl(cp->serial_number);
+#if defined(__Userspace__)
+#if defined(FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION)
+  serial_num = (asoc->asconf_seq_in + 1);
+#endif
+#endif
        if (SCTP_TSN_GE(asoc->asconf_seq_in, serial_num)) {
                /* got a duplicate ASCONF */
                SCTPDBG(SCTP_DEBUG_ASCONF1,
```

This makes it possible to send ASCONF chunks from the fuzzer without needing to fix up the sequence number to match what is expected by the receiver. When you hit an issue you can then generate the correct sequence numbers needed to reproduce an issue against an unmodified target, if desired.

Let me know if I should create a pull request for this?

tuexen:
I'm not sure (since I'm not a fuzzing expert) if this is a good idea. It forces that only a specific code path can be taken. So we'll test deterministically always some paths and never some others. If I remember it correctly from some bugs @weinrank found in the past, it did not take that long for the fuzzer to figure out how to modify the injected packet to take a code path. This resulted in behaviour where the peer would know secrets...

So I would expect the fuzzer (since it is coverage based, if my understanding is correct) to figure out by itself how to take interesting code paths.

BTW: Do you have any focus in testing usrsctp or are you just testing all available functionality?

markwo:
Thx - yes, valid points.  If the sequence number is not the expected value, the function returns without processing the ASCONF, so forcing it down the specific code path is actually needed if we want to cover this code.

With the approach @weinrank  is taking ("predictable randomness" during fuzzing), valid ASCONF sequence numbers are more likely to be found by the fuzzer. (And his fuzzer only sends one fuzzed packet per testcase, so the sequence number is also more likely to be found). So this change may not be strictly needed for your fuzzing code. I'd still argue it will make the fuzzer more effective (less wasted cycles when the sequence number is not the expected value).

I would suggest @weinrank checks the code coverage he gets when running the fuzzer for a while (e.g. overnight) and see if the ASCONF processing is being hit as expected, or the sequence number is an issue. I measured coverage using this approach with good results: https://github.com/google/fuzzer-test-suite/blob/master/tutorial/libFuzzerTutorial.md#visualizing-coverage

markwo:
Oh, and in terms of my focus: I'm just testing all functionality (trying to maximize code coverage during fuzzing)

PR #393, opened 2019-10-01
Fixed ticks overflow handling in timer implementation.

mstyura:
This fixes timer overflow and undefined behavior. It was observed on long running server that after `2^31` ticks since `usrsctp` initialization integer overflow happen.
`2^31` ticks with default time frequency `1000 Hz` (`1 tick` = `1 ms`) overflowed after `~24.8` days, causing server misbehavior.
~~With `uint64_t` used for ticks and default frequency of `1000 Hz` time can overflow only after `~584` billion years.~~
Proper handling of unsigned 32-bit integer overflow was implemented.

### DIFF
--- a/programs/st_client.c
+++ b/programs/st_client.c
@@ -98,7 +98,7 @@ handle_events(int sock, struct socket* s, void* sconn_addr)
 	int wait_time;
 
 	while (!finish) {
-		if (now > next_fire_time) {
+		if ((now - next_fire_time) & (1ULL<<63)) {
 			usrsctp_handle_timers(now - last_fire_time);
 			last_fire_time = now;
 			next_fire_time = now + TIMER_INTERVAL_MSECS;
--- a/programs/st_client.c
+++ b/programs/st_client.c
@@ -67,14 +67,14 @@
 static int connecting = 0;
 static int finish = 0;
 
-static uint64_t
+static unsigned int
 get_tick_count(void)
 {
 #ifdef _WIN32
-	return GetTickCount64();
+	return GetTickCount();
 #else
 	struct timeval tv;
-	uint64_t milliseconds;
+	unsigned int milliseconds;
 
 	gettimeofday(&tv, NULL); /* get current time */
 	milliseconds = tv.tv_sec*1000LL + tv.tv_usec/1000; /* calculate milliseconds */
@@ -92,13 +92,13 @@ handle_events(int sock, struct socket* s, void* sconn_addr)
 	fd_set rfds;
 	struct timeval tv;
 
-	uint64_t next_fire_time = get_tick_count();
-	uint64_t last_fire_time = next_fire_time;
-	uint64_t now = get_tick_count();
+	unsigned next_fire_time = get_tick_count();
+	unsigned last_fire_time = next_fire_time;
+	unsigned now = get_tick_count();
 	int wait_time;
 
 	while (!finish) {
-		if ((now - next_fire_time) & (1ULL<<63)) {
+		if ((int) (now - next_fire_time) > 0) {
 			usrsctp_handle_timers(now - last_fire_time);
 			last_fire_time = now;
 			next_fire_time = now + TIMER_INTERVAL_MSECS;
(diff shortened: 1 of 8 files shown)


mstyura:
Hello, @tuexen.

I was studying an sctp timer implementation for later investigation of proper possible fix to #325 .
During that I've figured out a possible explanation for server misbehavior after `~24` days uptime I've observed several times during past `~5` months in my environment.
There is ticks overflow happen due to `int` type usage (which was `32` bit on `64 bit` OS, which is fine). I've made ticks counter explicit type `uint64_t`, which is big enough, unsigned and does not cause "undefined behavior" from `C` compiler point of view.
Could you please review this changes?

Thanks a lot in advance!

tuexen:
Can you explain why overflowing is a problem? It is just a wrap around. Can you point me to a place where this results in unexpected behaviour?

mstyura:
I believe this line `233` from snippet below works wrong when overflow happen:
https://github.com/sctplab/usrsctp/blob/e67c6f8b30b8a8dfab5f04a40fa14cabffebec7c/usrsctplib/netinet/sctp_callout.c#L232-L234

For example, let's assume `c_time` inside timer was set to `2^31 - 2`.

`ticks` value before execution of line below was `2^31-9`
https://github.com/sctplab/usrsctp/blob/e67c6f8b30b8a8dfab5f04a40fa14cabffebec7c/usrsctplib/netinet/sctp_callout.c#L230
 and after execution it became `-2^32` assuming `elapsed_ticks` is `10`.
So,
https://github.com/sctplab/usrsctp/blob/e67c6f8b30b8a8dfab5f04a40fa14cabffebec7c/usrsctplib/netinet/sctp_callout.c#L233
will be evaluated as `2^31-2 <= -2^32` into `false`, but actually timer should execute it's callback.

And even more, [integer overflow](https://en.wikipedia.org/wiki/Integer_overflow) which happen here 
https://github.com/sctplab/usrsctp/blob/e67c6f8b30b8a8dfab5f04a40fa14cabffebec7c/usrsctplib/netinet/sctp_callout.c#L230
is [undefined behavior](https://en.wikipedia.org/wiki/Undefined_behavior) according to [`C` standard](http://port70.net/~nsz/c/c99/n1256.html#3.4.3p3), so it's just luck that `rm -rf --no-preserve-root /` was not executed as a result of UB 😄 .

tuexen:
You are correct with the first issue you are bringing up. That should be fixed.

Regarding your "undefined behaviour": We use that a lot in the stack for all sorts of sequence numbers which are wrapping around. That works. So we just need to be careful with comparisons. We can use the same method as for sequence numbers... Then it should work.

mstyura:
After you question I've further investigated proper way to handle time overflows. Here is [great analysis](https://arduino.stackexchange.com/a/12588) of a problem. I've fixed PR to match recommendations.

> We use that a lot in the stack for all sorts of sequence numbers which are wrapping around. That works.

That's terribly wrong idea. You think you know what undefined behavior does for you, but when someone else use slightly different set of compiler settings or different compiler, one will get build which has subtle bugs in it! It's terribly wrong to rely on signed integer overflow. Timers are crucial part of `usrsctp` it must be implemented reliable. `int` is wrong type to count ticks. If you insist I could change `uint64_t` to `uint32_t`, but it **must** be unsigned.

Thanks a lot for review!

mstyura:
I've rolled back `uint64_t` usage and used `uint32_t` where `int` was used. Ticks comparison has been fixed as well. Now changes look a lot smaller :)

tuexen:
What I meant (sorry for not being precise enough) was:
1. change the type of ticks to uint32_t. (Which is what you did)
2. Replace `c->c_time <= ticks` by `SCTP_UINT32_GE(ticks, c->c_time)`

This is what we use in a lot of other places in the stack. It works on a wide variety of platforms and compilers. Of course we are using unsigned ints.
Could you use the macro from 2.?

mstyura:
Thanks a lot @tuexen! `SCTP_UINT32_GE` is what I wanted (didn't know it is predefined already).

tuexen:
> Thanks a lot @tuexen! `SCTP_UINT32_GE` is what I wanted (didn't know it is predefined already).

Thanks for your work! Since we handle overflows for sequence numbers, we have the appropriate macros...

Issue #390, opened 2019-09-29
Can't achieve decent throughput on localhost


jean-airoldie:
I can't manage to get decent localhost performance. In my application code I get about `~50Mb/sec` (tested on 2 machines). Moreover, I seems like most of the time I'm waiting for the sendv & recvv to unblock.

My benchmark is simply a socket bound to localhost trying to sendv messages in a tight loop to another thread calling recvv, also in a tight loop. Note that I'm using the library as a cffi, however I don't think this is caused by the overhead of the cffi call.

My experimental setup:
* Ubuntu 18.04 (thought I also tried on centos7)
* Clang 8.0.0
* Benchmarking using `iperf3` I get `~500Mb/s` SCTP throughput and `~10Gb/s` UDP throughput.
* Compiled usrsctp with cmake in `Release` mode with `-O3` c flag & statically linked.
* UDP encapsulation at 9899 (but I also tried with the SCTP transport)
* Using 4Mb sendspace, recvspace, `RCV_BUF` and `SND_BUF` size.
* Using one-to-one sockets in non-blocking mode (but I also tried blocking mode).
* Tried varying `sendv` message size
* Tried varying application recv buffer size (the one provided to the `recvv` call)
* Tried varying the congestion algo just in case.

My `sendv` calls looks like this:
```c
usrsctp_sendv(
  sock,
  data,
  data_len,
  NULL,
  0,
  sendv_info, // a zero'ed sctp_sendv_spa with all flags cleared
  sizeof(sctp_sendv_spa),
  0
);
```
And my `recvv` calls:
```c
usrsctp_recvv(
  sock,
  data,
  data_len,  // 4Mb
  addr,
  sizeof(sctp_storage),
  info,
  sizeof(sctp_rcvinfo),
  0,
  0
);
```
I tried profiling with perf but I didn't find anything that looked especially abnormal, probably due to the multi-threaded nature of the lib.

Surely there must be something crucial I forgot.

lgrahl:
Same as #245 (even though that went slightly off-topic)

ivanhxq:
I also meet the same issue and urgently needed to be resolved!

msvoelker:
I used the tsctp program to reproduce the issue.

    git clone https://github.com/sctplab/usrsctp
    mkdir usrsctp/build
    cd usrsctp/build
    ccmake .. (disable sctp_debug)
    cmake --build .
    sudo programs/tsctp -L 127.0.0.1

In a second terminal:

[...]

Issue #245, opened 2018-07-14
CPU usage and throughput in WebRTC environments


lgrahl:
I've seen a multitude of independent WebRTC data channel implementations and all of them that use usrsctp underneath show the following behaviour:

* Throughput is CPU bound. For a modern i7 the cap is roughly at 500 Mbit/s in RAWRTC.
* Only one CPU thread is being used. (On Firefox, it jumps between CPU threads if e10s is active but overall is even slower.)

While I can't rule it out, I doubt that the encryption/decryption of DTLS packets is the limiting factor here. Still, it needs to be tested at some point.

500 Mbit/s seems fairly low to me for a modern CPU. Any ideas what might be the bottleneck? Is it the complexity of the protocol? What can we do to tackle this? Any student available? :grin: 

(We briefly talked about this a while ago but I thought it might be a good idea to file this.)

feross:
Thank you for opening this issue. This has plagued users of my popular [`simple-peer`](https://github.com/feross/simple-peer) library and WebTorrent itself for a very long time, even leading Brave to disable WebRTC in their built-in torrent viewer extension which is powered by WebTorrent, preferring to just connect to traditional torrent peers. Keeping WebRTC connections turned on just uses too much CPU.

lgrahl:
@feross Yeah, the main issue here is available resources (e.g. work hours). Perhaps we should also file this towards the bug trackers of Mozilla and Chrome to make them aware of it.

nxrighthere:
Not excluding the protocol itself, from what I see in the code (UDP encapsulation) the problem sits even lower. The current approach is non-scalable traditional UDP with a very basic I/O model where sooner or later the kernel (across platforms) will just reach its logical limits due to huge overhead. You have no idea what happens under the hood of the kernel with a single `sendmsg` `recvmsg` / `WSASendTo` `WSARecvFrom` call. A lock contention, memory allocations, inefficient queues management, route lookup, tons of system calls, and many other stuff is the reason why scalability is dying slowly right there.

Solution? [Scalable I/O](http://info.iet.unipi.it/~luigi/netmap/).

The only way to achieve a high-throughput is to build an architecture with an appropriate design in mind. It requires to rethink traditional ways of wrapping UDP sockets.

lgrahl:
Thanks for that insight but let's not jump to conclusions before we have practically analysed where the CPU spends its work.

[...]

PR #394, opened 2019-10-02
WIP: Use OS provided monotonic clock in timer implementation.

mstyura:
Currently `usrsctp` when compiled with `__Userspace__` uses it's own mechanism to count elapsed time between some events.
Timer implementation count [number of ticks](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L62) elapsed since library loaded. An actual [tick duration](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_constants.h#L601) depend on a pre defined SCTP timer frequency called [`hZ`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/user_environment.c#L50) with default to `1000 Hz` in userspace.
Timer is designed such that number of elapsed ticks is externally provided.
There is a build-in `while (true)` loop [enabled by default](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/user_socket.c#L112), which is executed in dedicated [`SCTP Timer` thread](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L283) and updating `ticks` with [fixed elapsed ticks rate](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L286).
When dedicated timer thread is disabled it's up to a library user to define elapsed  ticks between timer updates by [`usrsctp_handle_timers`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/user_socket.c#L3531)/[`sctp_handle_tick`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L221).

I see several problems with timer's `ticks` update implementation inside [`user_sctp_timer_iterate`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L281):
1. It assumes that there is exactly [`TIMEOUT_INTERVAL`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L278) physical time elapses during [`Sleep`](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep)/[`nanosleep`](http://man7.org/linux/man-pages/man2/nanosleep.2.html) execution. That's not true, OS thread scheduler might resume thread execution with bigger delay than `TIMEOUT_INTERVAL`, and that's not something unusual. An actual time spend in sleep is depend on gazilion things happens inside OS...
2. Elapsed time counting based on [`ticks` increment based on `MSEC_TO_TICKS(TIMEOUT_INTERVAL)`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L299) does not take into account time spent in the [`sctp_handle_tick`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L221) itself. Time spent in `sctp_handle_tick` is completely non-deterministic, because function implementation locks mutex several times and [execute arbitrary code form timer's `c_func`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L242).

The solution to the stated problems would be usage of `OS`-provided high-resolution timers, luckily all major OSes implements monotonic clock in some way.

Here in this pull request I've replaced manual `ticks` counting on `Windows`, `Linux/Android`, `MacOSX/iOS` with polling system's monotonic timer.
Then time from OS provided timer is converted back to `SCTP`s ticks which resolution defined by [`hZ`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/user_environment.c#L50).

With this change I expect better stability of the library in scenarios when timers are frequently used (retransmission for example).

This timer's implementation enhancement might also have positive influence on other computations where [`sctp_get_tick_count`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L67) used. For example, [congestion control computations](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_cc_functions.c) might have better precision, which also might have an influence on issues like #390 and #245.

I've tried to track down git history to find out the motivation behind current implementation based on manual `ticks` increment, but unfortunately there is not enough history here on github, because [commit where `sctp_callout.c` was added](https://github.com/sctplab/usrsctp/blob/dd7292377ed995d68a7ce4b359d173c1d9724af9/usrsctplib/netinet/sctp_callout.c) already has `ticks` variable and manual update. I might be wrong suggesting this changes as is, because I don't have a understanding of motivation behind current implementation. I'd be happy to know the motivation behind current timer design if anyone saved such knowledge from the past.


mstyura:
Hello @tuexen,
Could you please review this changes? I believe they should have a positive influence on timer implementation.
Thanks a lot in advance!

tuexen:
I'm not sure we need high precision timers currently in the stack. Can you provide a scenario where the current precision is not god enough?

We also have a bug in the timer system which results in deadlocks. I would like to delay optimisations after the bug is fixed.

usrsctplib/netinet/sctp_callout.c
```diff
@@ -55,22 +58,88 @@
 #include <netinet/sctp_pcb.h>
 #endif
 
+
+#if defined(__APPLE__) || defined(__Userspace__)
```
tuexen: Is it intentional that this change affects all userspace implementations and the kernel implementation on Mac OS?

mstyura:
> I'm not sure we need high precision timers currently in the stack. Can you provide a scenario where the current precision is not god enough?

Currently on Windows I've observed `10 - 25 %` time divergence between actual elapsed time and as it's counted by timer via `ticks`. And I've checked only basic scenario when system is almost idle and plugged to power - so basically no resource starvation and sleep works as precise as it can. I think timer inaccuracy must be fixed, if you think it need to be fixed after deadlock I'm fine with this. But eventually manual time counting via `ticks` must be offloaded to system's monotonic clock whenever possible.

usrsctplib/netinet/sctp_callout.c
mstyura: I indented to modify only userspace implementation. Will fix that.

mstyura:
> We also have a bug in the timer system which results in deadlocks. I would like to delay optimisations after the bug is fixed.

@tuexen  I'm working on a fix for that bug. 

I see the problem with current implementation.
The list of pending timers stored in the un-ordered doubly linked list.
The iteration routine [`sctp_handle_tick`](https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L221) iterates over double linked list and examine if timer need to be fired.
When timer is fired an associated function can modify double linked list of other timers is arbitrary way, so it really hard to make further iteration continue where it's stopped.
It was attempted to overcome arbitrary list mutations to continue iteration by using workarounds like `sctp_os_timer_next` and `sctp_os_timer_current`, but they don't not work well as it's done in current master: deadlocks, infinite loops - all that due to bug somewhere in that workarounds.

I see that more suitable data structure can be used to store list of timers - binary heap (priority queue). This data structure is widely used for exactly same task on other platforms, for example in [`Java`'s `ScheduledThreadPoolExecutor`](https://github.com/openjdk/jdk/blob/6bab0f539fba8fb441697846347597b4a0ade428/src/java.base/share/classes/java/util/concurrent/ScheduledThreadPoolExecutor.java#L902).
I'm currently working on implementation of binary heap and integrating into SCTP to store pending timers. I'm quite optimistic regarding ability of this priority queue to fix deadlock and infinite loop. It might take several days to make it stable, but I'm working on it.
What do you think about this plan, does it make sense for you?

tuexen:
In my view using a binary head data structure instead of list will be a performance optimization, but I'm not sure it fixes the dealdock problem. @peterlei What do you think?

shiretu:
@tuexen, @peterlei: how important is for `sctp_os_timer_stop` to finish the job it has (removing the timer) in that exact call? can it be deferred? If that's the case, the callout can be simply marked as deleted (another define like `SCTP_CALLOUT_STOPPED` along side `SCTP_CALLOUT_ACTIVE` or `SCTP_CALLOUT_PENDING`). The `sctp_handle_tick` function can then remove it. This way, the only locks will be ever held will be the ones protecting the queue

mstyura:
> In my view using a binary head data structure instead of list will be a performance optimization, but I'm not sure it fixes the dealdock problem

I expect that once I've implement and integrate binary heap to store timers I'll be able to get rid of this global state:
https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L86-L90

Which I expect will fix deadlock.

tuexen:
> > In my view using a binary head data structure instead of list will be a performance optimization, but I'm not sure it fixes the dealdock problem
> 
> I expect that once I've implement and integrate binary heap to store timers I'll be able to get rid of this global state:
> 
> https://github.com/sctplab/usrsctp/blob/1d6b4c2d0982f13f1184e254cc2341e07e71e22b/usrsctplib/netinet/sctp_callout.c#L86-L90
> 
> Which I expect will fix deadlock.

Will leave it to @peterlei  to comment on that, since I haven't analysed the deadlock problem myself. But I guess if you have a PR, some people would be happy to test the patch.

shiretu:
That will not get rid of the deadlock, almost sure if it. It really boils down on 2 threads wanting to mutate a container holding elements. When one of the threads is expected to exit the scope it has and do other calls, all bets are off: that thread is no longer deadlock safe. To make it safe, the thread doing removal should be responsible should be the same as the one in charge doing the removals

shiretu:
The best approach IMO is to sweep and execute all fallouts, skipping the stopped ones. All the call outs marked stopped can be saved on a local variable list and then simply removed one by one. Timers going in and out will not happen often.

shiretu:
Also, if the solution I have proposed is adopted, it will also elegantly solve the timer removal from other timer removals. Not sure if that’s an actual case, but it basically unwinds all graph of calls linearly. Deadlocks will be mathematically impossible

mstyura:
Hello, @tuexen,
I've have a question regarding my attempt to rewrite timer queue form list to binary heap. What is the library policy regarding dynamic memory allocations? I mean may I use `malloc/free` to implement binary heap based on dynamically allocated array or I have to avoid memory allocations at all and implement binary heap based on kind of node structure embedded into timer itself (similar way as it is done now via [`TAILQ_ENTRY`](https://github.com/sctplab/usrsctp/blob/05bea46702687f26a81c41c3fb1fd1dd3d9c0aa1/usrsctplib/netinet/sctp_callout.h#L92))?

PR #397, opened 2019-10-07
#325: SCTP timer deadlock fix

mstyura:
This PR is proposed as a fix for #325 

#### Problem description

Deadlock in bug #325 happen due to way [`sctp_os_timer_stop`](https://github.com/sctplab/usrsctp/blob/1ab1d69ce8cbf20a4d0d1e41de6b5c11cd24da71/usrsctplib/netinet/sctp_callout.c#L162) is **implemented** and they way it is **used**.

With [changes introduced from `25-30 March 2019`](https://github.com/sctplab/usrsctp/compare/1de72ec158a133866cb630ea49b3a6d92af76071...dfe89db408f29b78d667b9698e36325286be27ed) **the behavior of the [`sctp_os_timer_stop`](https://github.com/sctplab/usrsctp/blob/1ab1d69ce8cbf20a4d0d1e41de6b5c11cd24da71/usrsctplib/netinet/sctp_callout.c#L162) function has changed** in scenario when stop is called, but timer's **callback is still executing** at that moment.

**Before changes** from March 2019 stop was just noop in that case, because there nothing you can do to stop callback from running in other thread.

**With changes** from March 2019 stop in that case **became blocking function** which waits until callback of the timer finishes it's execution.

Unfortunately they way [`sctp_os_timer_stop`](https://github.com/sctplab/usrsctp/blob/1ab1d69ce8cbf20a4d0d1e41de6b5c11cd24da71/usrsctplib/netinet/sctp_callout.c#L162) is used by callers **has not been correspondingly updated** to take into account **new blocking behavior of stop** function.
As far as I understand blocking stop is necessary in cases when timer's memory is going to be recycled after stop, basically caller want to stop timer and free it's memory and make sure no one from timer's callback or timer's thread will touch already freed memory.
In `usrsctp` there is single macro and many usages of it where blocking behavior of stop is definitely needed, because the memory of the timer is freed just after stop:
https://github.com/sctplab/usrsctp/blob/1ab1d69ce8cbf20a4d0d1e41de6b5c11cd24da71/usrsctplib/netinet/sctp_var.h#L191-L211

There are a lot of places in `usrsctp` library where blocking behavior of stop is completely **not necessary** and actually causing deadlocks like #325.
All that places actually just want to **cancel** a timer, they don't need to **wait for completion** of the timer.

The deadlock as it happen in `usrsctp` due to blocking stop usage can be visualized with simple example `C++` program:
```C++
#include <iostream>
#include <mutex>
#include <future>
#include <thread>

int main(void)
{
	// Simulates some code which calls to sctp_os_timer_stop
	std::mutex mutex;
	std::lock_guard<std::mutex> guard(mutex);
	std::cout << "Non-timer thread [tid=" << std::this_thread::get_id() << "] will call to sctp_os_timer_stop, while holding mutex" << std::endl;

	{
		// Simulates timer
		std::future<void> timer = std::async(std::launch::async, [&mutex]()
		{
			// Simulates code executing on timer thread by sctp_handle_tick
			std::cout << "Timer thread [tid=" << std::this_thread::get_id() << "] callback is before acquiring mutex" << std::endl;
			std::lock_guard<std::mutex> guard_inside_timer_callback(mutex);
			std::cout << "Timer thread [tid=" << std::this_thread::get_id() << "] callback is after acquiring mutex" << std::endl;
		});

	} /* <-- exiting scope will cause recycle std::future memory by calling destructor,
		 but std::future must wait for completion before that, so causing deadlock,
		 the same story as with timer in usrsctp */

	std::cout << "Never printed due to deadlock with timer's callback function" << std::endl;

	return 0;
}
```

#### Existing implementation description
Brief description of an existing implementation:
1. Timers are stored in the single doubly linked list;
2. Timers inspection routine [`sctp_handle_tick`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L221) does linear scan of timers list and invokes triggered timer during iteration;
3. During iteration over timers list it is protected by special `mutex`, but it is released when timer's callback is invoked. **During execution of timer callback the state of doubly linked list can be arbitrary modified by other threads or by timer's callback itself**;
4. To not to scan timers list from scratch (i.e. resume) after single timer callback executed, `sctp_handle_tick` does some tricky bookkeeping and synchronization with [`sctp_os_timer_start`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L109) and [`sctp_os_timer_stop`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L162) based on:
[`sctp_os_timer_next`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L86);
4. To provide blocking functionality of stop to **wait for timer completion** the following global state maintained [`sctp_os_timer_current `](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L87), [`sctp_os_timer_waiting`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L88), [`sctp_os_timer_wait_ctr`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L89), [`sctp_os_timer_current_tid`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L90), [`sctp_os_timerwait_mtx`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L97), [`sctp_os_timer_wait_cond`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L98), [`sctp_os_timer_done_ctr`](https://github.com/sctplab/usrsctp/blob/c38740ec838e701f994b5885a1e7318b93de4843/usrsctplib/netinet/sctp_callout.c#L99).

#### Proposed implementation description
Brief description of proposed implementation:
1. Timers are stored in a linked binary heap (not very usual thing, usually binary heap is done based on dynamically allocated arrays. But linked heap implementation allow to avoid usage of explicitly allocated memory, so it's up to a caller where to store heap nodes).
2. During iteration and heap modification it's protected by same mutex as before;
3. `sctp_handle_tick` can no longer bother about heap modified arbitrary, because the state of the heap is maintained in such a way that soonest timer is always stored on top. So, the extra state and tricky synchronization between iteration, start and stop is no longer necessary, because `sctp_handle_tick` only check the top of the heap.
4. `sctp_os_timer_stop` behavior is left blocking, it does cancel of active timer and **waits until timer's callback finished execution**. `sctp_os_timer_stop` method still dangerous and has to be used carefully, but in scenarios where timer memory is need to be freed there is no other choice.
5. I've redefined several existing macroses and updated it's usage:
https://github.com/mstyura/usrsctp/blob/755599e1b226512ff0da7b8b9ad2963fd9bce6b9/usrsctplib/netinet/sctp_callout.h#L145-L149
* * `SCTP_OS_TIMER_STOP` - is defined to be cancel - the same as it was before [changes from `25-30 March 2019`](https://github.com/sctplab/usrsctp/compare/1de72ec158a133866cb630ea49b3a6d92af76071...dfe89db408f29b78d667b9698e36325286be27ed);
* * `SCTP_OS_TIMER_STOP_DRAIN` - is defined to be blocking, [`sctp_free_remote_addr`](https://github.com/sctplab/usrsctp/blob/1ab1d69ce8cbf20a4d0d1e41de6b5c11cd24da71/usrsctplib/netinet/sctp_var.h#L191-L211) macro which recycle timer memory defined to use blocking stop.
6. I've added stress test of `usrsctp` which reproduces deadlock #325. It was done as [CL to WebRTC project](https://webrtc-review.googlesource.com/c/src/+/157107). It is occurred it is a way easier to write examples using high level WebRTC API to reproduce issue in non-production environment.

This PR is likely to solve some of the deadlocks caused by usage of `sctp_os_timer_stop`, but it might be the case that there are still misuses of `sctp_os_timer_stop` and mutexes as shown in `C++` example - those cases need to be examined later once reproduced.

I'm really sorry that this PR is large and someone will have to review it.

### DIFF
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -423,6 +423,7 @@ user_sctp_timer_iterate(void *arg)
 		}
 		sctp_handle_tick(MSEC_TO_TICKS(TIMEOUT_INTERVAL));
 	}
+	sctp_userland_cond_destroy(&sctp_os_timer_current_completed);
 	return (NULL);
 }
 
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -349,8 +349,7 @@ sctp_handle_tick(uint32_t elapsed_ticks)
 	sctp_binary_heap_node_t* node = NULL;
 	while (0 == sctp_binary_heap_peek(heap, &node)) {
 		sctp_os_timer_t* c = (sctp_os_timer_t*)node->data;
-		if (!SCTP_UINT32_GE(ticks, c->c_time))
-		{
+		if (!SCTP_UINT32_GE(ticks, c->c_time)) {
 			if (last_heap_modification_reported != heap->mod_count) {
 				SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": the next soonest callout %p is scheduled at %" PRIu32 ", total scheduled callouts %zu\n",
 					__func__, ticks, c, c->c_time, sctp_binary_heap_size(heap));
@@ -385,7 +384,7 @@ sctp_handle_tick(uint32_t elapsed_ticks)
 		} else {
 			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": skipping callout %p with wrong flags %d\n", 
 				__func__, ticks, c, c->c_flags);
-			KASSERT(0, ("Timer from queue expected to be scheduled and active"));
+			KASSERT(0, ("Timer from queue expected to be scheduled and active and not cancelled"));
 		}
 	}
 	SCTP_TIMERQ_UNLOCK();
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -56,6 +56,15 @@
 #include <netinet/sctp_pcb.h>
 #endif
 
+/*
+ * Callout/Timer routines for OS that doesn't have them
+ */
+#if defined(__APPLE__) || defined(__Userspace__)
+static uint32_t ticks = 0;
+#else
+extern int ticks;
+#endif
+
 #if defined(__Userspace__)
 static void
 sctp_userland_cond_wait(userland_cond_t* cond,
@@ -130,19 +139,15 @@ sctp_os_timer_cancel_impl(sctp_os_timer_t* c) {
 	{
 		c->c_flags &= ~SCTP_CALLOUT_PENDING;
 		sctp_binary_heap_remove(&SCTP_BASE_INFO(timers_queue), &c->heap_node);
+		SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": cancelled pending callout %p\n",
+			__func__, ticks);
 		return (1);
 	}
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": cancelled non-pending callout %p\n",
+		__func__, ticks);
 	return (0);
 }
 
-/*
- * Callout/Timer routines for OS that doesn't have them
- */
-#if defined(__APPLE__) || defined(__Userspace__)
-static uint32_t ticks = 0;
-#else
-extern int ticks;
-#endif
 
 uint32_t sctp_get_tick_count(void) {
 	uint32_t ret;
@@ -172,6 +177,8 @@ sctp_os_timer_compare(const sctp_os_timer_t* a, const sctp_os_timer_t* b) {
 void
 sctp_os_timer_init(sctp_os_timer_t *c)
 {
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": request to init callout %p\n", 
+		__func__, sctp_get_tick_count(), c);
 	memset(c, 0, sizeof(*c));
 	sctp_binary_heap_node_init(&c->heap_node, c);
 #if defined(__Userspace__)
@@ -182,6 +189,8 @@ sctp_os_timer_init(sctp_os_timer_t *c)
 void
 sctp_os_timer_deinit(sctp_os_timer_t* c)
 {
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": request to deinit callout %p\n", 
+		__func__, sctp_get_tick_count(), c);
 	KASSERT(!sctp_os_timer_is_active(c), ("Deiniting an active timer"));
 #if defined(__Userspace__)
 	sctp_userland_cond_destroy(&c->c_completion);
@@ -207,6 +216,8 @@ sctp_os_timer_is_active(const sctp_os_timer_t* c) {
 void
 sctp_os_timer_deactivate(sctp_os_timer_t* c) {
 	SCTP_TIMERQ_LOCK();
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": request to deactivate callout %p\n", 
+		__func__, ticks, c);
 	c->c_flags &= ~SCTP_CALLOUT_ACTIVE;
 	SCTP_TIMERQ_UNLOCK();
 }
@@ -223,6 +234,8 @@ sctp_os_timer_start(sctp_os_timer_t *c, uint32_t to_ticks, void (*ftn) (void *),
 	}
 
 	SCTP_TIMERQ_LOCK();
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": request to start timer %p with delay of %" PRIu32 " ticks\n", 
+		__func__, ticks, c, to_ticks);
 	/* check to see if we're rescheduling a timer */
 	if ((c->c_flags & SCTP_CALLOUT_EXECUTING) != 0) {
 		/*
@@ -234,10 +247,14 @@ sctp_os_timer_start(sctp_os_timer_t *c, uint32_t to_ticks, void (*ftn) (void *),
 			 * This callout is already being stopped.
 			 * callout.  Don't reschedule.
 			 */
+			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": callout %p is already being stopped\n",
+				__func__, ticks, c);
 			SCTP_TIMERQ_UNLOCK();
 			return;
 		}
 	}
+
+	const uint32_t target_ticks = ticks + to_ticks;
 
 	sctp_binary_heap_t* timers_queue = &SCTP_BASE_INFO(timers_queue);
 	if (c->c_flags & SCTP_CALLOUT_PENDING) {
@@ -248,6 +265,8 @@ sctp_os_timer_start(sctp_os_timer_t *c, uint32_t to_ticks, void (*ftn) (void *),
 		 * flags.  We don't bother since we are setting these
 		 * below and we still hold the lock.
 		 */
+		SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": rescheduling callout %p from %" PRIu32 " to %" PRIu32 "\n",
+			__func__, ticks, c, c->c_time, target_ticks);
 	}
 
 	/*
@@ -260,8 +279,10 @@ sctp_os_timer_start(sctp_os_timer_t *c, uint32_t to_ticks, void (*ftn) (void *),
 	c->c_arg = arg;
 	c->c_flags = (SCTP_CALLOUT_ACTIVE | SCTP_CALLOUT_PENDING);
 	c->c_func = ftn;
-	c->c_time = ticks + to_ticks;
+	c->c_time = target_ticks;
 	sctp_binary_heap_push(timers_queue, &c->heap_node);
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": callout %p is scheduled with ticks %" PRIu32 "\n", 
+		__func__, ticks, c, target_ticks);
 	SCTP_TIMERQ_UNLOCK();
 }
 
@@ -269,6 +290,8 @@ int
 sctp_os_timer_cancel(sctp_os_timer_t* c) {
 	int ret;
 	SCTP_TIMERQ_LOCK();
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": request to cancel callout %p\n",
+		__func__, ticks, c);
 	ret = sctp_os_timer_cancel_impl(c);
 	SCTP_TIMERQ_UNLOCK();
 	return ret;
@@ -278,7 +301,8 @@ int
 sctp_os_timer_stop(sctp_os_timer_t *c)
 {
 	SCTP_TIMERQ_LOCK();
-
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": request to stop callout %p\n",
+		__func__, ticks, c);
 	sctp_os_timer_cancel_impl(c);
 
 	while (c->c_flags & SCTP_CALLOUT_EXECUTING) {
@@ -289,11 +313,18 @@ sctp_os_timer_stop(sctp_os_timer_t *c)
 		userland_thread_id_t tid;
 		sctp_userspace_thread_id(&tid);
 		if (sctp_userspace_thread_equal(tid, c->c_executor_id)) {
+			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": self cancel %p\n",
+				__func__, ticks, c);
 			break;
 		}
 #if defined(__Userspace__)
 		sctp_userland_cond_wait(&c->c_completion, &SCTP_BASE_VAR(timer_mtx));
 #endif
+
+		if (!(c->c_flags & SCTP_CALLOUT_EXECUTING)) {
+			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": waited callout completion %p\n", 
+				__func__, ticks, c);
+		}
 	}
 
 	SCTP_TIMERQ_UNLOCK();
@@ -303,16 +334,24 @@ sctp_os_timer_stop(sctp_os_timer_t *c)
 void
 sctp_handle_tick(uint32_t elapsed_ticks)
 {
+	static uint32_t last_heap_modification_reported = 0;
+
 	SCTP_TIMERQ_LOCK();
 	/* update our tick count */
 	ticks += elapsed_ticks;
 
+	uint32_t timedout_callouts = 0;
 	sctp_binary_heap_t* heap = &SCTP_BASE_INFO(timers_queue);
 	sctp_binary_heap_node_t* node = NULL;
 	while (0 == sctp_binary_heap_peek(heap, &node)) {
 		sctp_os_timer_t* c = (sctp_os_timer_t*)node->data;
 		if (!SCTP_UINT32_GE(ticks, c->c_time))
 		{
+			if (last_heap_modification_reported != heap->mod_count) {
+				SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": the next soonest callout %p is scheduled at %" PRIu32 "\n",
+					__func__, ticks, c, c->c_time);
+				last_heap_modification_reported = heap->mod_count;
+			}
 			// Earliest timer is not ready yet 
 			break;
 		}
@@ -328,14 +367,20 @@ sctp_handle_tick(uint32_t elapsed_ticks)
 			c->c_flags &= ~SCTP_CALLOUT_PENDING;
 
 			c->c_flags |= SCTP_CALLOUT_EXECUTING;
+			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": callout %p with to_ticks = %" PRIu32 " is about to execute\n", 
+				__func__, ticks, c, c->c_time);
 			SCTP_TIMERQ_UNLOCK();
 			c_func(c_arg);
 			SCTP_TIMERQ_LOCK();
+			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": callout %p with to_ticks = %" PRIu32 " is executed\n", 
+				__func__, ticks, c, c->c_time);
 			c->c_flags &= ~SCTP_CALLOUT_EXECUTING;
 #if defined(__Userspace__)
 			sctp_userland_cond_signal(&c->c_completion);
 #endif
 		} else {
+			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": skipping callout %p with wrong flags %d\n", 
+				__func__, ticks, c, c->c_flags);
 			KASSERT(0, ("Timer from queue expected to be scheduled and active"));
 		}
 	}
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -158,6 +158,7 @@ sctp_os_timer_wait_completion_impl(sctp_os_timer_t* c)
 			if (current_tid == c->c_executor_id)
 			{
 				// callout tried to wait for completion of itself
+				KASSERT(0, ("Deadlock detected: wait for self completion"));
 				retry_condition_wait = 0;
 			}
 			else
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -185,7 +185,7 @@ sctp_os_timer_start(sctp_os_timer_t *c, uint32_t to_ticks, void (*ftn) (void *),
 	case SCTP_CALLOUT_SCHEDULED:
 	{
 		sctp_binary_heap_remove(&SCTP_BASE_INFO(timers_queue), &c->heap_node);
-		break;
+		// FALLTHROUGH
 	}
 	case SCTP_CALLOUT_RUNNING:
 	case SCTP_CALLOUT_NEW:
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -339,7 +339,7 @@ sctp_os_timer_stop(sctp_os_timer_t *c)
 void
 sctp_handle_tick(uint32_t elapsed_ticks)
 {
-	static uint32_t last_heap_modification_reported = 0;
+	static uint32_t last_heap_version_reported = 0;
 
 	SCTP_TIMERQ_LOCK();
 	/* update our tick count */
@@ -350,10 +350,10 @@ sctp_handle_tick(uint32_t elapsed_ticks)
 	while (0 == sctp_binary_heap_peek(heap, &node)) {
 		sctp_os_timer_t* c = (sctp_os_timer_t*)node->data;
 		if (!SCTP_UINT32_GE(ticks, c->c_time)) {
-			if (last_heap_modification_reported != heap->mod_count) {
+			if (last_heap_version_reported != sctp_binary_heap_version(heap)) {
 				SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": the next soonest callout %p is scheduled at %" PRIu32 ", total scheduled callouts %zu\n",
 					__func__, ticks, c, c->c_time, sctp_binary_heap_size(heap));
-				last_heap_modification_reported = heap->mod_count;
+				last_heap_version_reported = sctp_binary_heap_version(heap);
 			}
 			// Earliest timer is not ready yet 
 			break;
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -46,6 +46,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <errno.h>
+#include <inttypes.h>
 #include <user_atomic.h>
 #include <netinet/sctp_sysctl.h>
 #include <netinet/sctp_pcb.h>
--- a/usrsctplib/netinet/sctp_callout.c
+++ b/usrsctplib/netinet/sctp_callout.c
@@ -46,7 +46,6 @@
 #include <string.h>
 #include <stdio.h>
 #include <errno.h>
-#include <inttypes.h>
 #include <user_atomic.h>
 #include <netinet/sctp_sysctl.h>
 #include <netinet/sctp_pcb.h>
@@ -55,136 +54,6 @@
 #include <netinet/sctp_callout.h>
 #include <netinet/sctp_pcb.h>
 #endif
-
-
-#if defined (__Userspace__)
-static
-void sctp_userland_cond_wait(userland_cond_t* cond, userland_mutex_t* mtx)
-{
-#if defined (__Userspace_os_Windows)
-	SleepConditionVariableCS(cond, mtx, INFINITE);
-#else	
-	int rc = pthread_cond_wait(cond, mtx);
-	if (rc)
-		SCTP_PRINTF("ERROR; return code from pthread_cond_wait is %d\n", rc);
-#endif
-}
-
-
-static
-void sctp_userland_cond_signal(userland_cond_t* cond)
-{
-#if defined (__Userspace_os_Windows)
-	WakeAllConditionVariable(cond);
-#else	
-	int rc = pthread_cond_broadcast(cond);
-	if (rc)
-		SCTP_PRINTF("ERROR; return code from pthread_cond_broadcast is %d\n", rc);
-#endif
-}
-
-static
-void sctp_userland_cond_init(userland_cond_t* cond)
-{
-#if defined (__Userspace_os_Windows)
-	InitializeConditionVariable(cond);
-#else	
-	int rc = pthread_cond_init(cond, NULL);
-	if (rc)
-		SCTP_PRINTF("ERROR; return code from pthread_cond_init is %d\n", rc);
-#endif
-}
-
-static
-void sctp_userland_cond_destroy(userland_cond_t* cond)
-{
-#if defined (__Userspace_os_Windows)
-	DeleteConditionVariable(cond);
-#else	
-	int rc = pthread_cond_destroy(cond);
-	if (rc)
-		SCTP_PRINTF("ERROR; return code from pthread_cond_destroy is %d\n", rc);
-#endif
-}
-#endif
-
-
-static void
-sctp_os_timer_cancel_impl(sctp_os_timer_t* c)
-{
-	switch (c->c_state)
-	{
-	case SCTP_CALLOUT_SCHEDULED:
-	{
-		sctp_binary_heap_remove(&SCTP_BASE_INFO(timers_queue), &c->heap_node);
-		c->c_state = SCTP_CALLOUT_COMPLETED;
-		break;
-	}
-	case SCTP_CALLOUT_RUNNING:
-	{
-		c->c_state = SCTP_CALLOUT_CANCEL_REQUESTED;
-		break;
-	}
-	case SCTP_CALLOUT_CANCEL_REQUESTED:
-	case SCTP_CALLOUT_COMPLETED:
-	case SCTP_CALLOUT_NEW:
-	{
-		// nothing to do
-		break;
-	}
-	default:
-	{
-		KASSERT(0, ("Unknown callout state"));
-		break;
-	}
-	}
-}
-
-
-static void
-sctp_os_timer_wait_completion_impl(sctp_os_timer_t* c)
-{
-	userland_thread_id_t current_tid;
-	sctp_userspace_thread_id(&current_tid);
-
-	int retry_condition_wait = 1;
-	do
-	{
-		switch (c->c_state)
-		{
-		case SCTP_CALLOUT_RUNNING:
-		case SCTP_CALLOUT_CANCEL_REQUESTED:
-		case SCTP_CALLOUT_SCHEDULED:
-		{
-			if (sctp_userspace_thread_equal(current_tid, c->c_executor_id))
-			{
-				// callout tried to wait for completion of itself
-				KASSERT(0, ("Deadlock detected: wait for self completion"));
-				retry_condition_wait = 0;
-			}
-			else
-			{
-#if defined (__Userspace__)
-				sctp_userland_cond_wait(&c->c_completion, &SCTP_BASE_VAR(timer_mtx));
-#endif
-			}
-			break;
-		}
-		case SCTP_CALLOUT_COMPLETED:
-		case SCTP_CALLOUT_NEW:
-		{
-			retry_condition_wait = 0;
-			break;
-		}
-		default:
-		{
-			KASSERT(0, ("Unknown callout state"));
-			break;
-		}
-		}
-	} while (retry_condition_wait);
-}
-
 
 /*
  * Callout/Timer routines for OS that doesn't have them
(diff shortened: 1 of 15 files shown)


mstyura:
Hello @tuexen.
Could you please review this PR? I know that at least some code style adjustment to match rest of code base are required before merge, but I've also might broken some assumptions existed in current implementation. I'd really like to hear feedback and eventually make this PR merged.
Thank you very much in advance!

mstyura:
Ok, looks like one crucial thing I've corrupted: I've removed ability to wait for completion inside stop (cancel) function. Will fix that.

mstyura:
Providing stop functionality which will not just cancel pending timer, but will wait until callback completion is non-trivial. Both `.NET` and `Java` has similar functionality provided via [`Task<T>`](https://github.com/dotnet/corefx/blob/7138b45be3c8df7e76629f7cc8b7ee5231ff7a56/src/Common/src/CoreLib/System/Threading/Tasks/Task.cs#L117) class and [`FutureTask<V>`](https://github.com/openjdk/jdk/blob/d00fd45b66baee07d8a03f36fee326c2c9da7d91/src/java.base/share/classes/java/util/concurrent/FutureTask.java#L66) correspondingly. I'll study their implementation to find out how to implement them for `usrsctp`. As for now it looks like currently `sctp_os_timer_stop` does two independent things and it might be better to split this method into 2: `sctp_os_timer_cancel` and `sctp_os_timer_wait_completion`.  So draining timer (cancelation with await for completion) could be implemented very naturally.

usrsctplib/netinet/sctp_callout.c
```diff
@@ -114,154 +174,197 @@ sctp_os_timer_start(sctp_os_timer_t *c, int to_ticks, void (*ftn) (void *),
...
+	} while (retry_condition_wait);
+
+	SCTP_TIMERQ_UNLOCK();
+}
+
+
+int
+sctp_os_timer_stop(sctp_os_timer_t *c)
+{
+	SCTP_TIMERQ_LOCK();
```
markwo: Is recursive locking enabled for this, on all platforms? From a quick look it seemed like it's not.

mstyura: Thank you very much! I've fixed that by extracting several helper functions which relies assumes that mutex is held by holder.

mstyura:
Hello, @tuexen!
I think now I have a PR in a state which fixes original deadlock preserving blocking stop where necessary. I've also added a test to repro an issue as [CL to WebRTC `AllMessagesAreDeliveredOverLoosyConnectionConcurrentTests`](https://webrtc-review.googlesource.com/c/src/+/157107). It is occurred that it's much easier to implement repro using WebRTC high level `API` rather than `usrsctp` directly, so that's why it's done as a CL.
Could you please review this pull request? Thank you very much!

mstyura:
Moving to WIP again, because I've found an fundamental issue with current integration:
https://github.com/mstyura/usrsctp/blob/6ecda929c17a4e1b7771f33d4a308f6dddb83f85/usrsctplib/netinet/sctp_callout.c#L380-L383
It's not possible to touch timer's memory here, because timer's memory might be already freed (thanks to ASAN). All that global variables now make sense for me...

mstyura:
Ok, now use after free inside timer is fixed by bringing back some of the global state (never de-referenced pointer to current timer and conditional variable to wait and signal current timer change). But there is a different problem happen *sometimes* [during test run](https://webrtc-review.googlesource.com/c/src/+/157107): `usrsctp` calls socket's [send callback `OnSctpOutboundPacket`](https://cs.chromium.org/chromium/src/third_party/webrtc/media/sctp/sctp_transport.cc?l=280&rcl=3663ed3ad67893746efd86267a936a7295031ab1) when socket [was already closed](https://cs.chromium.org/chromium/src/third_party/webrtc/media/sctp/sctp_transport.cc?l=433&rcl=3663ed3ad67893746efd86267a936a7295031ab1), this results in crash in [`SctpTransport` from `WebRTC`](https://cs.chromium.org/chromium/src/third_party/webrtc/media/sctp/sctp_transport.cc?rcl=3663ed3ad67893746efd86267a936a7295031ab1) due to usage of deleted object:
```
(gdb) bt
#0  0x000055555702d8d6 in rtc::AsyncInvoker::DoInvoke (this=0x61600109dbe0, posted_from=..., thread=0x0, closure=std::unique_ptr<class rtc::AsyncClosure> = 0x0 <nullptr>, id=0)
    at ../../rtc_base/async_invoker.cc:86
#1  0x0000555556efbd7a in rtc::AsyncInvoker::AsyncInvoke<void, rtc::MethodFunctor<cricket::SctpTransport, void (cricket::SctpTransport::*)(rtc::CopyOnWriteBuffer const&), void, rtc::CopyOnWriteBuffer const&> > (this=0x61600109dbe0, posted_from=..., thread=<optimized out>, functor=..., id=0) at ../../rtc_base/async_invoker.h:105
#2  cricket::SctpTransport::UsrSctpWrapper::OnSctpOutboundPacket (addr=<optimized out>, data=<optimized out>, length=<optimized out>, tos=<optimized out>, set_df=<optimized out>)
    at ../../media/sctp/sctp_transport.cc:296
#3  0x0000555557730787 in sctp_lowlevel_chunk_output (inp=<optimized out>, stcb=<optimized out>, net=<optimized out>, to=<optimized out>, m=0x61100863c500, 
    auth_offset=<optimized out>, auth=0x0, auth_keyid=0, nofragment_flag=1, ecn_ok=0, out_of_asoc_ok=<optimized out>, src_port=53824, dest_port=54080, v_tag=1159009962, 
    port=<optimized out>, over_addr=<optimized out>, so_locked=<optimized out>) at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_output.c:5043
#4  0x00005555577529e5 in sctp_send_abort_tcb (stcb=0x61e002271080, operr=<optimized out>, so_locked=<optimized out>)
    at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_output.c:11570
#5  0x00005555576eaeb0 in sctp_inpcb_free (inp=<optimized out>, immediate=<optimized out>, from=1) at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_pcb.c:4120
#6  0x00005555577f3139 in sctp_close (so=0x61600109dd80) at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_usrreq.c:910
#7  0x00005555576cf324 in sofree (so=0x61600109dd80) at ../../third_party/usrsctp/usrsctplib/usrsctplib/user_socket.c:298
#8  0x000055555770e051 in sctp_timeout_handler (t=<optimized out>) at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctputil.c:2081
#9  0x00005555577e0454 in sctp_handle_tick (elapsed_ticks=<optimized out>) at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_callout.c:372
#10 0x00005555577e0ac5 in user_sctp_timer_iterate (arg=<optimized out>) at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_callout.c:424
#11 0x00007ffff7f724cf in start_thread () from /usr/lib/libpthread.so.0
#12 0x00007ffff7bf42d3 in clone () from /usr/lib/libc.so.6
```
I'm currently trying to investigate whether this bug happen due to my changes or it was before them. I was not able to reproduce it on a [`usrsctp` version](https://github.com/sctplab/usrsctp/commit/7a8bc9a90ca96634aa56ee712856d97f27d903f8) used [currently by WebRTC](https://webrtc.googlesource.com/src/+/712b676e80f32b25d1e6d7f6eb2e49802e04a3bc/DEPS#234) - or I was not lucky enough to reproduce it.

As I can see from current debugging session the `sctp_handle_tick` triggers execution of `SCTP_TIMER_TYPE_HEARTBEAT` which decides to close socket and invoke send callback:
```
(gdb) frame 9
#9  0x00005555577e0454 in sctp_handle_tick (elapsed_ticks=<optimized out>) at ../../third_party/usrsctp/usrsctplib/usrsctplib/netinet/sctp_callout.c:372
372				c_func(c_arg);
(gdb) info locals
c_func = 0x55555770cf90 <sctp_timeout_handler>
c_time = <optimized out>
c_arg = 0x617000a33b90
c = <optimized out>
last_heap_version_reported = 27085970
node = 0x617000a33b90
heap = <optimized out>
(gdb) p *((struct sctp_timer*)node->data)
$5 = {
    timer = {
        heap_node = {
            data = 0x617000a33b90, 
            parent = 0x61700068f3f0, 
            left = 0x0, 
            right = 0x0, 
            heap = 0x55555895ca60 <system_base_info+808>, 
            sequence = 27086382}, 
        c_time = 710309,
        c_arg = 0x617000a33b90,
        c_func = 0x55555770cf90 <sctp_timeout_handler>, 
        c_flags = 6}, // 0b110 (SCTP_CALLOUT_ACTIVE | SCTP_CALLOUT_PENDING) 
    type = 5,  // SCTP_TIMER_TYPE_HEARTBEAT
    ep = 0x619000e93980,
    tcb = 0x61e002271080, 
    net = 0x617000a33b00, 
    self = 0x617000a33b90, 
    ticks = 679690,
    stopped_from = 0
}
```
It looks like heartbeat timer was not properly cancelled/drained during [previous invocation so socket close](https://cs.chromium.org/chromium/src/third_party/webrtc/media/sctp/sctp_transport.cc?l=836&rcl=3663ed3ad67893746efd86267a936a7295031ab1).

mstyura:
It is occurred that [`usrsctp_close`](https://github.com/sctplab/usrsctp/blob/09768bfcf22f91f6e43f0e692fc612564621d02f/usrsctplib/user_socket.c#L2175) might not dispose/free socket. `usrsctp_close` only remove reference from "user's" code and if there are any references to `struct socket *` the socket will survive `usrsctp_close` invocation:
https://github.com/sctplab/usrsctp/blob/09768bfcf22f91f6e43f0e692fc612564621d02f/usrsctplib/user_socket.c#L2192-L2194
where [`sorele`](https://github.com/sctplab/usrsctp/blob/09768bfcf22f91f6e43f0e692fc612564621d02f/usrsctplib/user_socketvar.h#L485-L495):
https://github.com/sctplab/usrsctp/blob/09768bfcf22f91f6e43f0e692fc612564621d02f/usrsctplib/user_socketvar.h#L485-L495
so, `sofree` is only invoked when last reference removed.
I've verified this by locally modifying `usrsctp_close` to have this:
```c
if (so->so_count != 1) 
{
	KASSERT(0, ("usrsctp_close will released socket, but there are more references"));
}
sorele(so);
```
And I've got this assert triggered:
```
(gdb) bt
#0  0x00007ffff7b30f25 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff7b1a897 in abort () from /usr/lib/libc.so.6
#2  0x00005555576ce91e in terminate_non_graceful () at ../../third_party/usrsctp/usrsctplib/usrsctplib/user_environment.h:94
#3  0x00005555576d7d0b in usrsctp_close (so=<optimized out>) at ../../third_party/usrsctp/usrsctplib/usrsctplib/user_socket.c:2190
#4  0x0000555556eefb5a in cricket::SctpTransport::CloseSctpSocket (this=0x616000708980) at ../../media/sctp/sctp_transport.cc:844
#5  cricket::SctpTransport::~SctpTransport (this=0x616000708980) at ../../media/sctp/sctp_transport.cc:435
#6  0x0000555556eefdce in cricket::SctpTransport::~SctpTransport (this=0x616000708980) at ../../media/sctp/sctp_transport.cc:433
#7  0x000055555666a3f6 in std::__1::default_delete<cricket::SctpTransport>::operator() (__ptr=0x2, this=<optimized out>)
    at ../../buildtools/third_party/libc++/trunk/include/memory:2338
#8  std::__1::unique_ptr<cricket::SctpTransport, std::__1::default_delete<cricket::SctpTransport> >::reset (this=0x60c000318728, __p=0x0)
    at ../../buildtools/third_party/libc++/trunk/include/memory:2651
#9  (anonymous namespace)::SctpPingPong::~SctpPingPong()::{lambda()#3}::operator()() const (this=<optimized out>) at ../../media/sctp/sctp_transport_reliability_unittest.cc:304
#10 rtc::FunctorMessageHandler<void, (anonymous namespace)::SctpPingPong::~SctpPingPong()::{lambda()#3}>::OnMessage(rtc::Message*) (this=<optimized out>, msg=0x7fffe96c2700)
    at ../../rtc_base/message_handler.h:60
#11 0x00005555570413a9 in rtc::MessageQueue::Dispatch (this=<optimized out>, pmsg=<optimized out>) at ../../rtc_base/message_queue.cc:514
#12 0x000055555706fdbb in rtc::Thread::ReceiveSendsFromThread (this=<optimized out>, source=<optimized out>) at ../../rtc_base/thread.cc:449
#13 0x000055555703f5a2 in rtc::MessageQueue::Get (this=0x6130000011c0, pmsg=<optimized out>, cmsWait=<optimized out>, process_io=<optimized out>)
    at ../../rtc_base/message_queue.cc:262
#14 0x000055555706f04c in rtc::Thread::ProcessMessages (this=<optimized out>, cmsLoop=<optimized out>) at ../../rtc_base/thread.cc:526
#15 0x000055555706eaea in rtc::Thread::PreRun (pv=0x6130000011c0) at ../../rtc_base/thread.cc:341
#16 0x00007ffff7f724cf in start_thread () from /usr/lib/libpthread.so.0
#17 0x00007ffff7bf42d3 in clone () from /usr/lib/libc.so.6
```
while `WebRTC` assume there will [no calls from `usrsctp` to `WebRTC`](https://cs.chromium.org/chromium/src/third_party/webrtc/media/sctp/sctp_transport.cc?l=836&rcl=3663ed3ad67893746efd86267a936a7295031ab1):
```c++
void SctpTransport::CloseSctpSocket() {
  RTC_DCHECK_RUN_ON(network_thread_);
  if (sock_) {
    // We assume that SO_LINGER option is set to close the association when
    // close is called. This means that any pending packets in usrsctp will be
    // discarded instead of being sent.
    usrsctp_close(sock_);
    sock_ = nullptr;
    usrsctp_deregister_address(this);
    UsrSctpWrapper::DecrementUsrSctpUsageCount();
    ready_to_send_data_ = false;
  }
}
```
with the assumption [repeated here](https://cs.chromium.org/chromium/src/third_party/webrtc/media/sctp/sctp_transport.cc?l=772&rcl=3663ed3ad67893746efd86267a936a7295031ab1):
```c++
// This ensures that the usrsctp close call deletes the association. This
// prevents usrsctp from calling OnSctpOutboundPacket with references to
// this class as the address.
linger linger_opt;
linger_opt.l_onoff = 1;
linger_opt.l_linger = 0;
if (usrsctp_setsockopt(sock_, SOL_SOCKET, SO_LINGER, &linger_opt,
                       sizeof(linger_opt))) {
    RTC_LOG_ERRNO(LS_ERROR) << debug_name_ << "->ConfigureSctpSocket(): "
                            << "Failed to set SO_LINGER.";
    return false;
}
```
I think there are several possible causes of the issue:
1.  `WebRTC`'s assumption was never correct and everything worked fine so far due to quite rare repro of the issue 
2. Between [`WebRTC`'s version](https://github.com/sctplab/usrsctp/commit/7a8bc9a90ca96634aa56ee712856d97f27d903f8) of `usrsctp` and current [`usrsctp` master](https://github.com/sctplab/usrsctp/commit/09768bfcf22f91f6e43f0e692fc612564621d02f) an extra reference [was introduced](https://github.com/sctplab/usrsctp/compare/7a8bc9a90ca96634aa56ee712856d97f27d903f8...09768bfcf22f91f6e43f0e692fc612564621d02f), which was not properly deleted from somewhere.

I don't know what is more likely. I'll wait for review from `usrsctp` maintainers and hope they will give more info.
For now removing WIP badge from PR.

mstyura:
Hello @tillz! Thank you very much for having a look at this PR and trying it in your environment! I've just fixed the issue you are mentioned. It looks like it would be better to destroy conditional variable when `SCTP timer` exits - so the function will not be unused and conditional variable will be properly recycled. Could you please let me know if you have other issues with this PR or it fixes deadlock for you let me know as well please. Thanks a lot!

tillz:
Thanks for your quick response!
Well, I'm definitely not having the expertise to comment on the code itself, however I can confirm that with your fixes my test program is running for ~40 minutes at full throttle (approx. 10 MB/s full duplex, CMT-wise split over two paths), whereas before it always locked up in less than a minute. So for me (as a user) it seems to work fine so far. Thank you very much for investigating this issue!

mstyura:
I've managed to reproduce `SIGSEGV` inside [`WebRTC` test](https://webrtc-review.googlesource.com/c/src/+/157107) in branch without my changes (mentioned [here](https://github.com/sctplab/usrsctp/pull/397#issuecomment-544484758)). I've reproduced it on [current `usrsctp` master ](https://github.com/sctplab/usrsctp/commit/09768bfcf22f91f6e43f0e692fc612564621d02f) with [commits introduced deadlock reverted](https://github.com/sctplab/usrsctp/compare/1de72ec158a133866cb630ea49b3a6d92af76071...dfe89db408f29b78d667b9698e36325286be27ed) on top of master. Will try bisect an issue...

mstyura:
@tuexen I'm really sorry to bother you again, but could you please have a look at this PR and provide any feedback on the work? It might be the case you completely disagree with fix, so that's why it ignored, so just let me know I could fix it in a different way. Thanks a lot in advance!

usrsctplib/netinet/sctp_var.h
```diff
@@ -239,9 +239,9 @@ extern struct pr_usrreqs sctp_usrreqs;
 #define sctp_free_remote_addr(__net) { \
 	if ((__net)) { \
 		if (SCTP_DECREMENT_AND_CHECK_REFCOUNT(&(__net)->ref_count)) { \
-			(void)SCTP_OS_TIMER_STOP(&(__net)->rxt_timer.timer); \
-			(void)SCTP_OS_TIMER_STOP(&(__net)->pmtu_timer.timer); \
-			(void)SCTP_OS_TIMER_STOP(&(__net)->hb_timer.timer); \
+			(void)SCTP_OS_TIMER_STOP_DRAIN(&(__net)->rxt_timer.timer); \
+			(void)SCTP_OS_TIMER_STOP_DRAIN(&(__net)->pmtu_timer.timer); \
+			(void)SCTP_OS_TIMER_STOP_DRAIN(&(__net)->hb_timer.timer); \
```
mstyura: Note for reviewers: 
I believe it is correct to use `SCTP_OS_TIMER_STOP_DRAIN`, because memory related to timers is freed just few lines below:
https://github.com/mstyura/usrsctp/blob/0e811d3eaa146a5fb3f58535638daaec41af2da9/usrsctplib/netinet/sctp_var.h#L255
where `SCTP_ZONE_FREE`:
https://github.com/mstyura/usrsctp/blob/0e811d3eaa146a5fb3f58535638daaec41af2da9/usrsctplib/netinet/sctp_os_userspace.h#L722-L725

usrsctplib/netinet/sctputil.c
```diff
@@ -1703,7 +1704,7 @@ sctp_timeout_handler(void *t)
 	}
 	type = tmr->type;
 	tmr->stopped_from = 0xa005;
-	SCTPDBG(SCTP_DEBUG_TIMER1, "Timer type %d goes off\n", type);
+	SCTPDBG(SCTP_DEBUG_TIMER1, "Timer %p of type %d goes off with callout is %p\n", tmr, type, &(tmr->timer));
```
mstyura: Note for reviewers: During development of deadlock fix I've made several stupid mistakes, so having more detailed logs related to timer proven to be helpful to find and fix those mistakes - here the relationship between `struct sctp_timer *` and `struct sctp_callout *` is explicitly logged.

usrsctplib/netinet/sctp_callout.h
```diff
@@ -99,20 +87,26 @@ typedef struct sctp_callout sctp_os_timer_t;
...
+int sctp_os_timer_cancel(sctp_os_timer_t*);
 int sctp_os_timer_stop(sctp_os_timer_t *);
 void sctp_handle_tick(uint32_t);
 
 #define SCTP_OS_TIMER_INIT	sctp_os_timer_init
 #define SCTP_OS_TIMER_START	sctp_os_timer_start
-#define SCTP_OS_TIMER_STOP	sctp_os_timer_stop
-/* MT FIXME: Is the following correct? */
-#define SCTP_OS_TIMER_STOP_DRAIN SCTP_OS_TIMER_STOP
-#define	SCTP_OS_TIMER_PENDING(tmr) ((tmr)->c_flags & SCTP_CALLOUT_PENDING)
```
mstyura: Note for reviewers:
I've found out that `c_flags` was read and mutated inconsistently: it mostly mutated under mutex `&SCTP_BASE_VAR(timer_mtx)`, but was read in many places directly. I believe this could be a reason of some timer related bugs.
I've decided to move reading `c_flags` under mutex and place it into dedicated function inside `sctp_callout.c` file.

#### The places where `c_flags` was read:
* ##### NOT holding `&SCTP_BASE_VAR(timer_mtx)` mutex:
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L2598
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L2629
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L2656
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L4299
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L4302
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L5201
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L5211
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_indata.c#L5216
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctputil.c#L1707
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctputil.c#L1747
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctputil.c#L1753
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctputil.c#L1759
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctputil.c#L2352
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L5113
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L8962
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L9264
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L9348
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L10355
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L10519
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L10604
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet6/sctp6_usrreq.c#L393
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_usrreq.c#L394
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_usrreq.c#L6186
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_usrreq.c#L6216
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_usrreq.c#L6334
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_output.c#L6368
* ##### Holding `&SCTP_BASE_VAR(timer_mtx)` mutex:
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L133
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L170
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L171
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L211
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L238

#### The places where `c_flags` was written:
* ##### NOT holding `&SCTP_BASE_VAR(timer_mtx)` mutex:
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctputil.c#L1759
* ##### holding mutex `&SCTP_BASE_VAR(timer_mtx)`
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L154
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L171
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L211
https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L238

usrsctplib/netinet/sctp_callout.h
```diff
@@ -99,20 +87,26 @@ typedef struct sctp_callout sctp_os_timer_t;
 
 #define	SCTP_CALLOUT_ACTIVE	0x0002	/* callout is currently active */
 #define	SCTP_CALLOUT_PENDING	0x0004	/* callout is waiting for timeout */
-
-void sctp_os_timer_init(sctp_os_timer_t *tmr);
+#define	SCTP_CALLOUT_CANCELLED	0x0008	/* callout is cancelled */
```
mstyura: Note for reviewers:
I've decided to replace cancellation flag of current timer implemented as global variable [`sctp_os_timer_waiting`](https://github.com/sctplab/usrsctp/blob/913de3599feded8322882bdae69f346da5a258fc/usrsctplib/netinet/sctp_callout.c#L88) with a state stored in timer itself.

usrsctplib/netinet/sctp_callout.c
```diff
@@ -75,72 +76,201 @@ uint32_t sctp_get_tick_count(void) {
...
+ * - SCTP_BASE_INFO(timers_queue)
  * - sctp_os_timer_current: current callout callback in progress
  * - sctp_os_timer_current_tid: current callout thread id in progress
- * - sctp_os_timer_waiting: some thread is waiting for callout to complete
- * - sctp_os_timer_wait_ctr: incremented every time a thread wants to wait
- *                           for a callout to complete.
+ * - sctp_os_timer_current_changed: conditional variable signaled when
+ *                                  current callout pointer is changed
  */
-static sctp_os_timer_t *sctp_os_timer_next = NULL;
```
mstyura: Note for reviewers:
It is no longer necessary to store pointer to next timer as global variable and keep it in sync when timer's list state changed (in `sctp_os_timer_start`/`sctp_os_timer_stop`/`sctp_handle_tick`). This is because linked binary heap is now used to store timers and it maintains state to provide next soonest timer to iterate over all expired timers.

It is still necessary to store never de-referenced pointer to current timer in `sctp_os_timer_current`. The pointer to current timer is used to determine currently running timer. This "executing" flag can not be stored in timer's state, because timer can self-detroy in callback and there is no way to reset "executing" flag in timer's memory, because it was already freed.

usrsctplib/netinet/sctp_callout.c
```diff
@@ -75,72 +76,201 @@ uint32_t sctp_get_tick_count(void) {
...
- * - sctp_os_timer_wait_cond: waiting for callout to complete
- * - sctp_os_timer_done_ctr: value of "wait_ctr" after triggering "waiting"
- */
-userland_mutex_t sctp_os_timerwait_mtx;
-static userland_cond_t sctp_os_timer_wait_cond;
-static int sctp_os_timer_done_ctr = 0;
+#if defined(__Userspace__)
+
+static void
+sctp_userland_cond_wait(userland_cond_t* cond,
```
mstyura: Note for reviewers:
I've preferred to extract set of helper functions to work with condtional variables. I prefer simple helper functions over macroses.

usrsctplib/netinet/sctp_callout.c
```diff
@@ -75,72 +76,201 @@ uint32_t sctp_get_tick_count(void) {
...
+                        userland_mutex_t* mtx) {
+	// callee must be robust to spurious wakeups, both because
+	// wakeup could happen due to native API behavior and because
+	// there is indendend timeout for wait.
+#if defined(__Userspace_os_Windows)
+	const DWORD timeoutMillis = 20 * 1000;
+	const BOOL waited = SleepConditionVariableCS(cond, mtx, timeoutMillis);
+	if (!waited) {
+		SCTP_PRINTF("WARN; SleepConditionVariableCS did not return within %ul millis\n", timeoutMillis);
```
mstyura: Note for reviewers:
I think this `printf` and non-infinite wait timeout could be useful for future debugging, whenever someone misuse blocking `sctp_os_timer_stop` when some mutex necessary to execute timer's callback is held (deadlock condition).

usrsctplib/netinet/sctp_callout.c
```diff
@@ -75,72 +76,201 @@ uint32_t sctp_get_tick_count(void) {
...
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": cancelled non-pending callout %p\n",
+		__func__, ticks, c);
+	return (0);
+}
+
+void
+sctp_os_timer_describe(const sctp_os_timer_t* t,
+                       size_t max_len,
+                       char* buffer) {
+	snprintf(buffer, max_len, "t=%" PRIu32 ",a=%p", t->c_time, (void*)t);
```
mstyura: Note for reviewers:
This function used as part of printing tree of timers. The presence of such function is really useful when debugging obscure timer behavior under debugger. For example, when debugging under `gdb` the current timer's queue state could be printed with:
`call sctp_binary_heap_print(system_base_info.sctppcbinfo.timers_queue)`

usrsctplib/netinet/sctp_callout.c
```diff
@@ -75,72 +76,201 @@ uint32_t sctp_get_tick_count(void) {
...
 			/*
 			 * This callout is already being stopped.
 			 * callout.  Don't reschedule.
 			 */
+			SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": callout %p is cancelled\n",
+				__func__, ticks, c);
 			SCTP_TIMERQ_UNLOCK();
 			return;
 		}
+		if ((c->c_flags & SCTP_CALLOUT_ACTIVE) == 0) {
```
mstyura: Note for reviewers:
This is not strictly necessary, but it is added as mean of extra debugging of timers

usrsctplib/netinet/sctp_callout.c
```diff
@@ -151,117 +281,115 @@ sctp_os_timer_start(sctp_os_timer_t *c, uint32_t to_ticks, void (*ftn) (void *),
...
-		c->c_flags &= ~SCTP_CALLOUT_ACTIVE;
-		if (sctp_os_timer_current != c) {
-			SCTP_TIMERQ_UNLOCK();
-			return (0);
-		} else {
+	SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": request to stop callout %p\n",
+		__func__, ticks, c);
+	sctp_os_timer_cancel_impl(c);
+
+	while (c == sctp_os_timer_current) {
```
mstyura: Note for reviewers:
There is a `sctp_os_timer_current_changed` conditional variable which is fired whenever `sctp_os_timer_current` is changed. The `sctp_os_timer_stop` must not finish until timer to stop is current timer (if it is not self-stop).

usrsctplib/netinet/sctp_callout.c
```diff
@@ -151,117 +281,115 @@ sctp_os_timer_start(sctp_os_timer_t *c, uint32_t to_ticks, void (*ftn) (void *),
...
-			TAILQ_REMOVE(&SCTP_BASE_INFO(callqueue), c, tqe);
-			c_func = c->c_func;
-			c_arg = c->c_arg;
+
+	sctp_binary_heap_t* heap = &SCTP_BASE_INFO(timers_queue);
+	sctp_binary_heap_node_t* node = NULL;
+	while (0 == sctp_binary_heap_peek(heap, &node)) {
+		sctp_os_timer_t* c = (sctp_os_timer_t*)node->data;
+		if (!SCTP_UINT32_GE(ticks, c->c_time)) {
+			if (last_heap_version_reported != sctp_binary_heap_version(heap)) {
```
mstyura: Note for reviewers:
Currently `sctp_handle_tick` is invoked `~10 ms`, which could result in polluted logs saying  `the next soonest callout ...`. Remembering last version printed allow not printing same message multiple times.
The further enhancement would be changing `sctp_handle_tick` invocation to be result of timed wait on event/conditional variable (this will be done as dedicated PR once this PR is merged).

usrsctplib/netinet/sctp_callout.h
```diff
@@ -52,44 +54,30 @@ __FBSDID("$FreeBSD$");
 
 #define SCTP_TICKS_PER_FASTTIMO 20	/* called about every 20ms */
 
-extern userland_mutex_t sctp_os_timerwait_mtx;
```
mstyura: Note for reviewers:
Because `sctp_os_timer_current_changed` was introduced and it is signaled whenever `sctp_os_timer_current` re-assigned it is no longer necessary to have dedicated mutex to handle blocking stop. The combination of `sctp_os_timer_current`, `sctp_os_timer_current_changed` and `&SCTP_BASE_VAR(timer_mtx)` is used to implement blocking stop.

usrsctplib/netinet/sctp_callout.c
```diff
@@ -75,72 +76,201 @@ uint32_t sctp_get_tick_count(void) {
...
+	int rc = pthread_cond_destroy(cond);
+	if (rc) {
+		SCTP_PRINTF("ERROR; return code from pthread_cond_destroy is %d\n", rc);
+	}
+#endif
+}
+#endif
+
+static int
+sctp_os_timer_cancel_impl(sctp_os_timer_t* c) {
```
mstyura: Note for reviewers:
This is private helper function used both by `stop` (blocking) and `cancel` (non-blocking) functions. It updates timer state update (`c_flags`) and ensure timer is not in `timers_queue` (removing it in queue).

usrsctplib/netinet/sctp_callout.c
```diff
@@ -75,72 +76,201 @@ uint32_t sctp_get_tick_count(void) {
...
+	// Assume SCTP_TIMERQ_LOCK done by caller
+
+	c->c_flags &= ~(SCTP_CALLOUT_ACTIVE);
+	c->c_flags |= (SCTP_CALLOUT_CANCELLED);
+
+	if (c->c_flags & SCTP_CALLOUT_PENDING)
+	{
+		c->c_flags &= ~SCTP_CALLOUT_PENDING;
+		sctp_binary_heap_remove(&SCTP_BASE_INFO(timers_queue), &c->heap_node);
+		SCTPDBG(SCTP_DEBUG_TIMER2, "%s: now=%" PRIu32 ": cancelled pending callout %p\n",
```
mstyura: Notes for reviewers:
`SCTPDBG(SCTP_DEBUG_TIMER2, ...)` has proven being invaluable source of information during fix development for debugging with logs.

mstyura:
Closing in favor of proper fix #417

taylor-b:
No need to abandon this entirely, it sounds like you found some issues with the timers unrelated to the deadlock which this fixes, no?

mstyura:
@taylor-b yes and I’m going to replace this big PR with set of small one, once your deadlock fix is in master. Does it sound  reasonable?

taylor-b:
Sounds good.

PR #400, opened 2019-10-09
Added initial support of compilation by clang-cl.exe on Windows.

mstyura:
`Visual Studio` has support of `CMake` projects with clang toolchain on `Windows`, but currently `usrsctp` cmake file passes clang-specific switches to `clang-cl.exe`, which expect the same switches format as `cl.exe`.
With proposed changes `clang-cl.exe` will be treated as `cl.exe` with small exception to extra warnings provided by clang.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,7 +161,7 @@ endif ()
 #################################################
 
 # Determine if compiler is Visual Studio compiler or Clang in MSVC compatible mode
-if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC" OR "${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
+if (CMAKE_C_COMPILER_ID MATCHES "MSVC" OR CMAKE_C_SIMULATE_ID MATCHES "MSVC")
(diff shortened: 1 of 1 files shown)


tuexen:
@weinrank: What do you think?

mstyura:
@tuexen ideally I'd like to fix warnings I've disabled when compiled with `clang-cl` (has done some work to fix them, but I'm not yet ready to share that work). Also I'd like to enable sanitizers and fuzzer on `Windows` as one of the means (among I'm considering) to test changes from #397. `Windows` in that case is more convenient to me because I have more experience in debugging with `VS` and `WinDBG` rather than `lldb` or `gdb` in case some problem is found during #397 tests and need to be carefully debugged.

weinrank:
Havent tested it but the code looks good!

mstyura:
I've checked with my private build bot that compilation on Windows, Linux, Android, iOS and MacOS X still works. On Windows changes are successfully compiled with both `clang-cl.exe` and `cl.exe`. Btw, is there any way to try your build bot before PR is merged? I see build bot only builds project after merge request is accepted and merged. Thanks a lot for review!

PR #402, opened 2019-11-08
meson: Fixes needed to build on Android

nirbheek:
commit 48fee387dd98748bd3b4ed4e1dd80b016a228b17:

```
meson: add support for android/ios host_system
```

commit 5f1a0f75b3406688d8ad95cf616e43244a9c820e:

```
build: include ifaddrs.c definition for android
```

With these changes libusrsctp builds correctly on Android with Meson.

CC: @ystreet

PR #403, opened 2019-11-08
meson: Add support for building on Android

nirbheek:
The ifaddrs API is a GNU extension which is not available with Android's bionic libc, so we need to ship our own implementation of it which calls the appropriate Linux kernel Netlink APIs.

The code is by Kenneth MacKay and is mirrored at https://github.com/oliviertilmans/android-ifaddrs

The license is 2-clause BSD.

The addition of these files was accidentally missed when I submitted https://github.com/sctplab/usrsctp/pull/402, because of which the changes from that PR don't actually work. Sorry about that!

### DIFF
--- a/usrsctplib/meson.build
+++ b/usrsctplib/meson.build
@@ -1,9 +1,9 @@
 # Includes
-include_dirs = include_directories(
+include_dirs = [include_directories(
     '.',
     'netinet',
     'netinet6',
-)
+)]
 
 # Sources
 sources = files([
@@ -13,7 +13,8 @@ sources = files([
     'user_socket.c',
 ])
 if system == 'android'
-    sources += files('ifaddrs.c')
+    sources += files('android/ifaddrs.c')
+    include_dirs += include_directories('android')
 endif
 
 subdir('netinet')
--- a/usrsctplib/android/ifaddrs.c
+++ b/usrsctplib/android/ifaddrs.c
@@ -1,0 +1,663 @@
+/*
+Copyright (c) 2013, Kenneth MacKay
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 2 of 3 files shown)


tuexen:
How does this related to the Android stuff Mozilla is using in https://github.com/mozilla/gecko-dev/blob/master/netwerk/sctp/src/ifaddrs_android.cpp?

nirbheek:
Seems like exactly the same thing implemented in C++ instead of C.

lgrahl:
It's completely unnecessary for Mozilla (even though it exists) because they don't use the code for WebRTC data channels.

@nirbheek If you're in a WebRTC environment, just configure it with `sctp_inet=false sctp_inet6=false`. If you're not, this is legit.

In any case, I do not understand why this line has been added: https://github.com/sctplab/usrsctp/pull/402#discussion_r345741637. IIRC it should be reverted.

nirbheek:
> @nirbheek If you're in a WebRTC environment, just configure it with `sctp_inet=false sctp_inet6=false`. If you're not, this is legit.

We use usrsctp for [data channels](https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/blob/master/ext/sctp/meson.build) in [GStreamer's WebRTC implementation](http://blog.nirbheek.in/2018/02/gstreamer-webrtc.html), so we do need this.

> In any case, I do not understand why this line has been added: [#402 (comment)](https://github.com/sctplab/usrsctp/pull/402#discussion_r345741637). IIRC it should be reverted.

Yes, that was my mistake. Sorry! That patch should've been what's in this PR. I can submit another PR to revert that if you'd like. Alternatively we could merge this PR which also fixes it :D

usrsctplib/meson.build
```diff
@@ -1,9 +1,9 @@
 # Includes
-include_dirs = include_directories(
+include_dirs = [include_directories(
```
lgrahl: I'm wondering if this isn't already an array. Have you tested it without the brackets?

nirbheek: Yes, I tested it without the brackets and that didn't work (even though I expected it to). Unlike `files()`, `include_directories()` doesn't implement the `+=` operator. I'm not sure why.

lgrahl:
> We use usrsctp for [data channels](https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/blob/master/ext/sctp/meson.build) in [GStreamer's WebRTC implementation](http://blog.nirbheek.in/2018/02/gstreamer-webrtc.html), so we do need this, unless I misunderstood something?

Then you don't need this patch and you can simply disable both `sctp_inet=false sctp_inet6=false` when configuring the Meson build which will disable the code paths that need the ifaddrs API. That is because you will need to wrap it in a DTLS packet anyway and can't operate on top of sockets directly.

nirbheek:
> That is because you will need to wrap it in a DTLS packet anyway and can't operate on top of sockets directly.

Hmm! That's a good point. I should've realized that. It does work just fine if I disable those options. We don't actually need this patch.

What do you want to do here? If you would like to close this PR, I'll open a new one that reverts the bad ifaddrs.c addition patch.

lgrahl:
I've opened #404 for the fix. :slightly_smiling_face:

usrsctplib/meson.build
nirbheek: I looked at the parser, and the reason is that `files()` returns an array of `File` objects, while `include_directories()` contains multiple include dirs in a single object. Could be fixed upstream.

nirbheek:
#404 merged, closing.

PR #413, opened 2019-12-16
sctp_input: Do not dereference a NULL pointer

sancane:


tuexen:
Can you reproduce a NULL pointer dereference?
My understanding is that if `sctp_findassociation_addr()` returns with `net != NULL`, the `stcb` is also not NULL. You can't find a path without finding an association this path belongs to. So if you run into a NULL pointer dereference, there is a bug in `sctp_findassociation_addr()`.

sancane:
Actually I did this patch when I was revising the code looking for fix for issue https://github.com/sctplab/usrsctp/issues/414. I just saw that function sctp_findassociation_addr could return NULL and there was no special treatment for it whereas a few lines later there was another call to the same fuction and the "if" statement did check for the case NULL before dereferencing the sctp pointer.

tuexen:
Since `net != NULL` implies `stcb != NULL`, this is not an issue in my view.

PR #417, opened 2020-01-09
Do proper reference counting with callouts

taylor-b:
This handles the race condition between a timer executing and being deleted.

sctp_timer_start increments reference counts, and the timer executing or sctp_timer_stop being called decrements them.

### DIFF
--- a/usrsctplib/netinet/sctp_pcb.c
+++ b/usrsctplib/netinet/sctp_pcb.c
@@ -4149,6 +4149,7 @@ sctp_inpcb_free(struct sctp_inpcb *inp, int immediate, int from)
 	if (SCTP_ASOC_CREATE_LOCK_CONTENDED(inp))
 		being_refed++;
 
+	/* NOTE: 0 refcount also means no timers are referencing us. */
 	if ((inp->refcount) ||
 	    (being_refed) ||
 	    (inp->sctp_flags & SCTP_PCB_FLAGS_CLOSE_IP)) {
@@ -4170,19 +4171,6 @@ sctp_inpcb_free(struct sctp_inpcb *inp, int immediate, int from)
 	SCTP_INP_WUNLOCK(inp);
 	SCTP_ASOC_CREATE_UNLOCK(inp);
 	SCTP_INP_INFO_WUNLOCK();
-	/* Now we release all locks. Since this INP
-	 * cannot be found anymore except possibly by the
-	 * kill timer that might be running. We call
-	 * the drain function here. It should hit the case
-	 * were it sees the ACTIVE flag cleared and exit
-	 * out freeing us to proceed and destroy everything.
-	 */
-	if (from != SCTP_CALLED_FROM_INPKILL_TIMER) {
-		(void)SCTP_OS_TIMER_STOP_DRAIN(&inp->sctp_ep.signature_change.timer);
-	} else {
-		/* Probably un-needed */
-		(void)SCTP_OS_TIMER_STOP(&inp->sctp_ep.signature_change.timer);
-	}
 
 #ifdef SCTP_LOG_CLOSING
 	sctp_log_closing(inp, NULL, 5);
--- a/usrsctplib/netinet/sctputil.c
+++ b/usrsctplib/netinet/sctputil.c
@@ -1673,8 +1673,14 @@ sctp_timeout_handler(void *t)
 	struct socket *upcall_socket = NULL;
 #endif
 	int did_output;
+	int released_asoc_reference;
 	int type;
 
+	/*
+	 * NOTE: If inp, stcb or net are non-null, then references to these were
+	 * added when the timer was started, and must be released before this
+	 * function returns.
+	 */
 	tmr = (struct sctp_timer *)t;
 	inp = (struct sctp_inpcb *)tmr->ep;
 	stcb = (struct sctp_tcb *)tmr->tcb;
@@ -1683,6 +1689,7 @@ sctp_timeout_handler(void *t)
 	CURVNET_SET((struct vnet *)tmr->vnet);
 #endif
 	did_output = 1;
+	released_asoc_reference = 0;
 
 #ifdef SCTP_AUDITING_ENABLED
 	sctp_audit_log(0xF0, (uint8_t) tmr->type);
@@ -1695,10 +1702,7 @@ sctp_timeout_handler(void *t)
 		 * SCTP_PRINTF("Stale SCTP timer fired (%p), ignoring...\n",
 		 *             (void *)tmr);
 		 */
-#if defined(__FreeBSD__) && __FreeBSD_version >= 801000
-		CURVNET_RESTORE();
-#endif
-		return;
+		goto out_decr;
 	}
 	tmr->stopped_from = 0xa001;
 	if (!SCTP_IS_TIMER_TYPE_VALID(tmr->type)) {
@@ -1706,22 +1710,15 @@ sctp_timeout_handler(void *t)
 		 * SCTP_PRINTF("SCTP timer fired with invalid type: 0x%x\n",
 		 * tmr->type);
 		 */
-#if defined(__FreeBSD__) && __FreeBSD_version >= 801000
-		CURVNET_RESTORE();
-#endif
-		return;
+		goto out_decr;
 	}
 	tmr->stopped_from = 0xa002;
 	if ((tmr->type != SCTP_TIMER_TYPE_ADDR_WQ) && (inp == NULL)) {
-#if defined(__FreeBSD__) && __FreeBSD_version >= 801000
-		CURVNET_RESTORE();
-#endif
-		return;
+		goto out_decr;
 	}
 	/* if this is an iterator timeout, get the struct and clear inp */
 	tmr->stopped_from = 0xa003;
 	if (inp) {
-		SCTP_INP_INCR_REF(inp);
 		if ((inp->sctp_socket == NULL) &&
 		    ((tmr->type != SCTP_TIMER_TYPE_INPKILL) &&
 		     (tmr->type != SCTP_TIMER_TYPE_INIT) &&
@@ -1732,58 +1729,33 @@ sctp_timeout_handler(void *t)
 		     (tmr->type != SCTP_TIMER_TYPE_SHUTDOWNACK) &&
 		     (tmr->type != SCTP_TIMER_TYPE_SHUTDOWNGUARD) &&
 		     (tmr->type != SCTP_TIMER_TYPE_ASOCKILL))) {
-			SCTP_INP_DECR_REF(inp);
-#if defined(__FreeBSD__) && __FreeBSD_version >= 801000
-			CURVNET_RESTORE();
-#endif
-			return;
+			goto out_decr;
 		}
 	}
 	tmr->stopped_from = 0xa004;
-	if (stcb) {
-		atomic_add_int(&stcb->asoc.refcnt, 1);
-		if (stcb->asoc.state == 0) {
-			atomic_add_int(&stcb->asoc.refcnt, -1);
-			if (inp) {
-				SCTP_INP_DECR_REF(inp);
-			}
-#if defined(__FreeBSD__) && __FreeBSD_version >= 801000
-			CURVNET_RESTORE();
-#endif
-			return;
-		}
+	if (stcb && stcb->asoc.state == 0) {
+		goto out_decr;
 	}
 	type = tmr->type;
 	tmr->stopped_from = 0xa005;
 	SCTPDBG(SCTP_DEBUG_TIMER1, "Timer type %d goes off\n", type);
 	if (!SCTP_OS_TIMER_ACTIVE(&tmr->timer)) {
-		if (inp) {
-			SCTP_INP_DECR_REF(inp);
-		}
-		if (stcb) {
-			atomic_add_int(&stcb->asoc.refcnt, -1);
-		}
-#if defined(__FreeBSD__) && __FreeBSD_version >= 801000
-		CURVNET_RESTORE();
-#endif
-		return;
+		goto out_decr;
 	}
 	tmr->stopped_from = 0xa006;
 
 	if (stcb) {
 		SCTP_TCB_LOCK(stcb);
+		/*
+		 * Release reference so that association can be freed if necessary below.
+		 * This is safe now that we have acquired the lock.
+		 */
 		atomic_add_int(&stcb->asoc.refcnt, -1);
+		released_asoc_reference = 1;
 		if ((type != SCTP_TIMER_TYPE_ASOCKILL) &&
 		    ((stcb->asoc.state == 0) ||
 		     (stcb->asoc.state & SCTP_STATE_ABOUT_TO_BE_FREED))) {
-			SCTP_TCB_UNLOCK(stcb);
-			if (inp) {
-				SCTP_INP_DECR_REF(inp);
-			}
-#if defined(__FreeBSD__) && __FreeBSD_version >= 801000
-			CURVNET_RESTORE();
-#endif
-			return;
+			goto get_out;
 		}
 	} else if (inp != NULL) {
 		if (type != SCTP_TIMER_TYPE_INPKILL) {
@@ -2132,8 +2104,15 @@ out_decr:
 		sorele(upcall_socket);
 	}
 #endif
+	/* Note that these reference counts were incremented in sctp_timer_start. */
 	if (inp) {
 		SCTP_INP_DECR_REF(inp);
+	}
+	if (stcb && !released_asoc_reference) {
+		atomic_add_int(&stcb->asoc.refcnt, -1);
+	}
+	if (net) {
+		sctp_free_remote_addr(net);
 	}
 
 out_no_decr:
@@ -2157,7 +2136,13 @@ sctp_timer_start(int t_type, struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 	if (stcb) {
 		SCTP_TCB_LOCK_ASSERT(stcb);
 	}
-	/* Don't restart timer on net that's been removed. */
+	/* Don't restart timer on association that's about to be killed. */
+	if (stcb != NULL &&
+	    (stcb->asoc.state & SCTP_STATE_ABOUT_TO_BE_FREED) &&
+	    (t_type != SCTP_TIMER_TYPE_ASOCKILL)) {
+		return;
+	}
+	/* Similarly, don't restart timer on net that's been removed. */
 	if (net != NULL && (net->dest_state & SCTP_ADDR_BEING_DELETED)) {
 		return;
 	}
@@ -2428,7 +2413,21 @@ sctp_timer_start(int t_type, struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 #ifndef __Panda__
 	tmr->ticks = sctp_get_tick_count();
 #endif
-	(void)SCTP_OS_TIMER_START(&tmr->timer, to_ticks, sctp_timeout_handler, tmr);
+	if (!SCTP_OS_TIMER_START(&tmr->timer, to_ticks, sctp_timeout_handler, tmr)) {
+		/*
+		 * If this is a newly scheduled callout, as opposed to a rescheduled one,
+		 * increment relevant reference counts.
+		 */
+		if (inp) {
+			SCTP_INP_INCR_REF(inp);
+		}
+		if (stcb) {
+			atomic_add_int(&stcb->asoc.refcnt, 1);
+		}
+		if (net) {
+			atomic_add_int(&net->ref_count, 1);
+		}
+	}
 	return;
 }
 
@@ -2579,7 +2578,21 @@ sctp_timer_stop(int t_type, struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 	}
 	tmr->self = NULL;
 	tmr->stopped_from = from;
-	(void)SCTP_OS_TIMER_STOP(&tmr->timer);
+	if (SCTP_OS_TIMER_STOP(&tmr->timer)) {
+		/*
+		 * If timer was actually stopped, decrement reference counts that were
+		 * incremented in sctp_timer_start.
+		 */
+		if (inp) {
+			SCTP_INP_DECR_REF(inp);
+		}
+		if (stcb) {
+			atomic_add_int(&stcb->asoc.refcnt, -1);
+		}
+		if (net) {
+			sctp_free_remote_addr(net);
+		}
+	}
 	return;
 }
 
@@ -8424,4 +8437,3 @@ sctp_add_substate(struct sctp_tcb *stcb, int substate)
 	}
 #endif
 }
-
--- a/usrsctplib/netinet/sctp_callout.h
+++ b/usrsctplib/netinet/sctp_callout.h
@@ -95,6 +95,11 @@ int sctp_os_timer_stop(sctp_os_timer_t *);
 void sctp_handle_tick(uint32_t);
 
 #define SCTP_OS_TIMER_INIT	sctp_os_timer_init
+/*
+ * NOTE: These shouldn't be called directly outside of sctp_timer_start and
+ * sctp_timer_stop, since they don't handle incrementing/decrementing
+ * relevant reference counts.
+ */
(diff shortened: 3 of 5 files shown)


taylor-b:
Note that it would be

usrsctplib/netinet/sctp_callout.c
```diff
@@ -77,60 +79,28 @@ uint32_t sctp_get_tick_count(void) {
...
  * - SCTP_BASE_INFO(callqueue)
  * - sctp_os_timer_next: next timer to check
- * - sctp_os_timer_current: current callout callback in progress
- * - sctp_os_timer_current_tid: current callout thread id in progress
- * - sctp_os_timer_waiting: some thread is waiting for callout to complete
- * - sctp_os_timer_wait_ctr: incremented every time a thread wants to wait
- *                           for a callout to complete.
  */
 static sctp_os_timer_t *sctp_os_timer_next = NULL;
-static sctp_os_timer_t *sctp_os_timer_current = NULL;
```
taylor-b: Note that this and a lot of stuff below is just reverting the problematic commits that cause the deadlock. Again, look at individual commits for clearer diff.

usrsctplib/netinet/sctp_callout.c
```diff
@@ -77,60 +79,28 @@ uint32_t sctp_get_tick_count(void) {
...
-
 
 void
 sctp_os_timer_init(sctp_os_timer_t *c)
 {
 	memset(c, 0, sizeof(*c));
 }
 
-void
+int
```
taylor-b: This return value is consistent with callout_reset, which is what SCTP_OS_TIMER_START maps to in FreeBSD.

usrsctplib/netinet/sctp_pcb.c
```diff
@@ -4151,10 +4149,10 @@ sctp_inpcb_free(struct sctp_inpcb *inp, int immediate, int from)
 	if (SCTP_ASOC_CREATE_LOCK_CONTENDED(inp))
 		being_refed++;
 
+	// NOTE: 0 refcount also means no timers are referencing us.
 	if ((inp->refcount) ||
 	    (being_refed) ||
 	    (inp->sctp_flags & SCTP_PCB_FLAGS_CLOSE_IP)) {
-		(void)SCTP_OS_TIMER_STOP(&inp->sctp_ep.signature_change.timer);
```
taylor-b: This was already stopped above, so I don't think it needs to be called again?

usrsctplib/netinet/sctp_pcb.c
```diff
@@ -4173,19 +4171,6 @@ sctp_inpcb_free(struct sctp_inpcb *inp, int immediate, int from)
 	SCTP_INP_WUNLOCK(inp);
 	SCTP_ASOC_CREATE_UNLOCK(inp);
 	SCTP_INP_INFO_WUNLOCK();
-	/* Now we release all locks. Since this INP
```
taylor-b: No longer necessary since at this point we are assured no timer referencing us is running.

usrsctplib/netinet/sctp_indata.c
```diff
@@ -1745,7 +1745,7 @@ sctp_process_a_data_chunk(struct sctp_tcb *stcb, struct sctp_association *asoc,
 		 * empty data chunk.
 		 */
 		op_err = sctp_generate_no_user_data_cause(tsn);
-		stcb->sctp_ep->last_abort_code = SCTP_FROM_SCTP_INDATA + SCTP_LOC_14;
+		stcb->sctp_ep->last_abort_code = SCTP_FROM_SCTP_INDATA + SCTP_LOC_15;
```
mstyura: @taylor-b is this change done to resolve uniqueness violation, because `SCTP_LOC_14` is used here:
https://github.com/sctplab/usrsctp/blob/32020d6f44d405ef4ea3fecb52bc8a392aa8a6c4/usrsctplib/netinet/sctp_indata.c#L1592
?

usrsctplib/netinet/sctp_pcb.c
mstyura: Note: the place where it stopped is here:

https://github.com/sctplab/usrsctp/blob/0f1de1e6896523ac790ca0880fd6c91dc8a3d7cb/usrsctplib/netinet/sctp_pcb.c#L3932-L3933

usrsctplib/netinet/sctp_pcb.c
```diff
@@ -5734,11 +5737,9 @@ sctp_free_assoc(struct sctp_inpcb *inp, struct sctp_tcb *stcb, int from_inpcbfre
 	 */
 	(void)SCTP_OS_TIMER_STOP(&asoc->strreset_timer.timer);
 	(void)SCTP_OS_TIMER_STOP(&asoc->dack_timer.timer);
-	(void)SCTP_OS_TIMER_STOP(&asoc->strreset_timer.timer);
```
mstyura: As far I understand here `SCTP_OS_TIMER_STOP` is used and not `sctp_timer_stop` because otherwise reference counter would be decremented twice, where first time it was decremented here:
https://github.com/sctplab/usrsctp/blob/0f1de1e6896523ac790ca0880fd6c91dc8a3d7cb/usrsctplib/netinet/sctp_pcb.c#L5555-L5568 is that correct?

usrsctplib/netinet/sctp_pcb.c
```diff
@@ -5533,36 +5534,38 @@ sctp_free_assoc(struct sctp_inpcb *inp, struct sctp_tcb *stcb, int from_inpcbfre
...
-		net->hb_timer.self = NULL;
+		sctp_timer_stop(SCTP_TIMER_TYPE_SEND, inp, stcb, net,
+		                SCTP_FROM_SCTP_PCB + SCTP_LOC_22);
+		sctp_timer_stop(SCTP_TIMER_TYPE_INIT, inp, stcb, net,
+		                SCTP_FROM_SCTP_PCB + SCTP_LOC_23);
+		sctp_timer_stop(SCTP_TIMER_TYPE_SHUTDOWN, inp, stcb, net,
+		                SCTP_FROM_SCTP_PCB + SCTP_LOC_24);
+		sctp_timer_stop(SCTP_TIMER_TYPE_COOKIE, inp, stcb, net,
+		                SCTP_FROM_SCTP_PCB + SCTP_LOC_25);
+		sctp_timer_stop(SCTP_TIMER_TYPE_SHUTDOWNACK, inp, stcb, net,
```
mstyura: As I understand `SCTP_TIMER_TYPE_SEND`, `SCTP_TIMER_TYPE_INIT`, `SCTP_TIMER_TYPE_SHUTDOWN`, `SCTP_TIMER_TYPE_COOKIE`, `SCTP_TIMER_TYPE_SHUTDOWNACK` all maps to `net->rxt_timer`, so it means that ref counter would be decremented 5 times. Is that correct that existing code already properly used `sctp_timer_start` which does increment?

usrsctplib/netinet/sctputil.c
```diff
@@ -2529,7 +2537,21 @@ sctp_timer_stop(int t_type, struct sctp_inpcb *inp, struct sctp_tcb *stcb,
...
+		 * incremented in sctp_timer_start.
+		 */
+		if (inp) {
+			SCTP_INP_DECR_REF(inp);
+		}
+		if (stcb) {
+			atomic_add_int(&stcb->asoc.refcnt, -1);
+		}
+		if (net) {
+			sctp_free_remote_addr(net);
```
mstyura: Not really sure about this, but better to clarify rather than ignore: shouldn't reference be decremented in the reverse order of being incremented?

usrsctplib/netinet/sctp_pcb.c
mstyura: BTW, other than this block of `SCTP_OS_TIMER_STOP` usage there one more usage here:

https://github.com/sctplab/usrsctp/blob/0f1de1e6896523ac790ca0880fd6c91dc8a3d7cb/usrsctplib/netinet/sctp_pcb.c#L8072

Shouldn't `sctp_timer_stop` be used here or at least comment why not?

usrsctplib/netinet/sctp_indata.c
taylor-b: Yes, as long as I was adding more SCTP_LOCs I went ahead and made the rest ordered and unique.

usrsctplib/netinet/sctp_pcb.c
taylor-b: That's correct. And yikes, can't believe I missed that, good catch. Fixed.

usrsctplib/netinet/sctp_pcb.c
taylor-b: It won't decrement five times, because the timer can only be in the list once, with one type, so only the call with a matching type will cause a decrement. It's unfortunate that one function call turned into five, but I didn't want to make a separate timer stop function that takes the timer as an argument just for this. Though maybe I should?

And yes, sctp_timer_start was already being called everywhere.

usrsctplib/netinet/sctputil.c
taylor-b: It doesn't really make a difference here, but I could change it for stylistic purposes if people want.

usrsctplib/netinet/sctp_pcb.c
mstyura: > Though maybe I should?

IMO, no. Just wanted to clarify the behavior here.

usrsctplib/netinet/sctputil.c
```diff
@@ -1686,58 +1732,33 @@ sctp_timeout_handler(void *t)
 		     (tmr->type != SCTP_TIMER_TYPE_SHUTDOWNACK) &&
 		     (tmr->type != SCTP_TIMER_TYPE_SHUTDOWNGUARD) &&
 		     (tmr->type != SCTP_TIMER_TYPE_ASOCKILL))) {
-			SCTP_INP_DECR_REF(inp);
```
mstyura: This decrement migrated to here as I understand:
https://github.com/sctplab/usrsctp/blob/4b77ee20ffa7a2b684f48e802f7b529a7151d978/usrsctplib/netinet/sctputil.c#L2584-L2598

mstyura:
Reviewed latest version of this PR. Looks good to me.

mstyura:
I've just realized that there is one more direct usage of `sctp_timer_stop` caused by usage of [`SCTP_OS_TIMER_STOP_DRAIN`](https://github.com/taylor-b/usrsctp/blob/4b77ee20ffa7a2b684f48e802f7b529a7151d978/usrsctplib/netinet/sctp_callout.h#L106). 
https://github.com/sctplab/usrsctp/blob/4b77ee20ffa7a2b684f48e802f7b529a7151d978/usrsctplib/netinet/sctp_pcb.c#L6943
which could possibly replaced with:
```
sctp_timer_stop(SCTP_TIMER_TYPE_ADDR_WQ,
            ???,
            ???,
            ???);
```
It will probably does not affect reference counting, but it might make sense to change it as well (and remove `SCTP_OS_TIMER_STOP_DRAIN` completely).

taylor-b:
@mstyura This usage is okay because it doesn't involve any reference counted structures. And given everything that's about to be freed I assumed the "stop drain" behavior is desired here.

mstyura:
@tuexen hello prof. Michael Tüxen!
Could you please review this pull request?
Thanks a lot in advance!

tuexen:
> @tuexen hello prof. Michael Tüxen!
> Could you please review this pull request?
> Thanks a lot in advance!

It is on my agenda for next week...

lgrahl:
It's great to see you back, @taylor-b!

tuexen:
Hi @taylor-b ,

is it OK for you if I cherry pick the cleanup changes from the PR, commit them, and then you rebase the patch? Then we can easily focus on non-cleanup changes.

taylor-b:
@tuexen Yeah, that's fine.

taylor-b:
@tuexen Rebased.

tuexen:
@taylor-b Can you rebase (again) and avoid comments of the form `// ...`. Please replace them ny using `/* ... */`. The remaining changes should be the actual fix.

taylor-b:
@tuexen Done.

tuexen:
> @tuexen Done.

Thanks. Will review and test it also in the FreeBSD kernel stack.

usrsctplib/netinet/sctp_pcb.c
```diff
@@ -5237,6 +5225,21 @@ sctp_remove_net(struct sctp_tcb *stcb, struct sctp_nets *net)
...
+	                SCTP_FROM_SCTP_PCB + SCTP_LOC_11);
+	sctp_timer_stop(SCTP_TIMER_TYPE_COOKIE, inp, stcb, net,
+	                SCTP_FROM_SCTP_PCB + SCTP_LOC_12);
+	sctp_timer_stop(SCTP_TIMER_TYPE_SHUTDOWNACK, inp, stcb, net,
+	                SCTP_FROM_SCTP_PCB + SCTP_LOC_13);
+	sctp_timer_stop(SCTP_TIMER_TYPE_PATHMTURAISE, inp, stcb, net,
+	                SCTP_FROM_SCTP_PCB + SCTP_LOC_14);
+	sctp_timer_stop(SCTP_TIMER_TYPE_HEARTBEAT, inp, stcb, net,
+	                SCTP_FROM_SCTP_PCB + SCTP_LOC_15);
+	net->removed = 1; /* Prevents timers from being restarted. */
```
tuexen: I think stopping the HB and PMTU timer here is actually the correct place. However, we should not stop the other timers here, since, if for example the SEND timer is not running anymore, no retransmissions for the DATA chunks sent on the now removed net will happen. I committed a corresponding fix.

usrsctplib/netinet/sctp_pcb.c
```diff
@@ -5531,35 +5534,8 @@ sctp_free_assoc(struct sctp_inpcb *inp, struct sctp_tcb *stcb, int from_inpcbfre
...
-	asoc->dack_timer.self = NULL;
-	(void)SCTP_OS_TIMER_STOP(&asoc->strreset_timer.timer);
-	/*-
-	 * For stream reset we don't blast this unless
-	 * it is a str-reset timer, it might be the
-	 * free-asoc timer which we DON'T want to
-	 * disturb.
-	 */
-	if (asoc->strreset_timer.type == SCTP_TIMER_TYPE_STRRESET)
-		asoc->strreset_timer.self = NULL;
```
tuexen: Since the assoc kill timer is kept in place, I added a boolean variable to `sctp_stop_association_timers()`.

tuexen:
I committed all the timer code cleanup to reduce the diff for the fix. I also took stopping the HB and PMTU timer an net removal. I think we can't stop the others...

Can you rebase? The diff should be much smaller now.

usrsctplib/netinet/sctp_pcb.c
taylor-b: Can you summarize how retransmissions work? I'm now worried that the `net->removed` logic is also flawed; should it be possible to restart the SEND timer after removing?

I'm doing this because the old behavior was "stop timers when there are zero references". This doesn't work since the timers themselves now add references.

So I assumed (pretty naively I now realize) that I could just stop the timers here. If that's not possible, then we still have the same issue: a timer holding a reference could keep the net alive when it otherwise would have been freed...

If we want exactly the same behavior as before, maybe we could use two reference counts as I mentioned previously.

When the main reference count is reduced to 0, we'd stop the timers (as they were stopped before), but would only free the memory if the timer reference count is also 0.

The new problem would be that we'd be setting the `removed`/"about to be freed" flag inside `sctp_free_remote_addr`, which would be racy since I'm assuming there is no guarantee that the association lock is held wherever that's called.

So I'm somewhat at a loss now. We could address that problem by adding a `NETKILL` timer but now things are getting pretty messy.

I went ahead and rebased, but I want to be sure this is watertight before merging. Do you have any ideas?

usrsctplib/netinet/sctp_callout.c
```diff
@@ -49,10 +49,12 @@
 #include <user_atomic.h>
 #include <netinet/sctp_sysctl.h>
 #include <netinet/sctp_pcb.h>
+#include <netinet/sctp_var.h>
```
tuexen: I think this is not needed. Can you remove it?

usrsctplib/netinet/sctp_callout.c
```diff
@@ -49,10 +49,12 @@
...
 #else
 #include <netinet/sctp_os.h>
 #include <netinet/sctp_callout.h>
 #include <netinet/sctp_pcb.h>
+#include <netinet/sctp_var.h>
```
tuexen: I think this is not needed. Can you remove it?

usrsctplib/netinet/sctp_pcb.c
tuexen: > Can you summarize how retransmissions work? I'm now worried that the `net->removed` logic is also flawed; should it be possible to restart the SEND timer after removing?

If you send a DATA chunk on a `net`, you store a pointer to it in `whoTo` and start a timer for it. Assuming we have multiple `net`s, when the timer runs off, we choose a different net, retransmit the DATA chunk there and start a new timer on the other `net`.

If we remove a `net`, and there is outstanding data, we need to wait until the timer runs off. In the old code, we would not free the net since there was a ref count on all outstanding DATA chunks. But I think this should work with the new stuff the same way.

I guess this can be optimized be moving chunks over to a different `net`, but that can be a separate commit.
This also does not apply to the WebRTC case, since dynamic address configuration is not used there.
> 
> I'm doing this because the old behavior was "stop timers when there are zero references". This doesn't work since the timers themselves now add references.
We would remove the net, after the timer is running off. That is the same as in the old case, since we would have ref counts on the DATA chunks.
> 
> So I assumed (pretty naively I now realize) that I could just stop the timers here. If that's not possible, then we still have the same issue: a timer holding a reference could keep the net alive when it otherwise would have been freed...

I don't see that as an issue and the behaviour should be the same.

> 
> If we want exactly the same behavior as before, maybe we could use two reference counts as I mentioned previously.
> 
> When the main reference count is reduced to 0, we'd stop the timers (as they were stopped before), but would only free the memory if the timer reference count is also 0.

When we are removing the `net`, it does not make any sense to send packets via the `net`. So we should stop the HB and PMTU timer right away. That is why I committed that. The DATA retransmission must be kept running... Or me move the data away...

> 
> The new problem would be that we'd be setting the `removed`/"about to be freed" flag inside `sctp_free_remote_addr`, which would be racy since I'm assuming there is no guarantee that the association lock is held wherever that's called.

I think the STCB lock should be hold when modifying any `net`. Do you have a place in mind, where it is not held?

> 
> So I'm somewhat at a loss now. We could address that problem by adding a `NETKILL` timer which would repeatedly attempt to stop the timers, but now things are getting pretty messy.
> 
> I went ahead and rebased, but I want to be sure this is watertight before merging. Do you have any ideas?

Thanks for re-basing. I'm trying to commit things in small pieces to reduce the diff and get a better view regarding the the changes. I think ref counting should work. I'm also putting the changes to the FreeBSD kernel, since we have there some features for detecting memory loss or locking issues...

tuexen:
I replaced the `bool` by an `int` as you suggested but used the `SCTP_ADDR_BEING_DELETED` flag of the `dest_state` field` instead of adding a `removed` field to the `net` structure. After the removal of `#include <netinet/sctp_var.h>`, could you rebase again?

usrsctplib/netinet/sctp_pcb.c
taylor-b: >  Do you have a place in mind, where it is not held?
No, I just wasn't sure and didn't look exhaustively. I assumed that since the atomic increment was being used, that was the only thing guaranteeing thread safety.

Thanks for your explanation and for bearing with me.

usrsctplib/netinet/sctp_callout.c
taylor-b: Done, and rebased.

taylor-b:
@tuexen Ping

tuexen:
@taylor-b Still working on some cleanup of the timer code...

tuexen:
@taylor-b Just to understand your concept crystal clear:

- Whenever you start a timer, increment refcounts to whatever is referenced (inp, stcb, net).
- Whenever you stop a running timer, decrement refcounts to whatever is referenced (inp, stcb, net).
- Whenever a timer runs off, decrement refcounts to whatever is referenced (inp, stcb, net).

Here is an overview of the various timers:

|Timer				|Object				|ref inp	|ref stcb	|ref net	|
|-------------------------------|-------------------------------|-------|-------|-------|
|SCTP_TIMER_TYPE_SEND		|net->rxt_timer			|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_INIT		|net->rxt_timer			|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_RECV		|stcb->asoc.dack_timer		|Yes	|Yes	|No	|
|SCTP_TIMER_TYPE_SHUTDOWN	|net->rxt_timer			|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_HEARTBEAT	|net->hb_timer			|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_COOKIE		|net->rxt_timer			|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_NEWCOOKIE	|inp->sctp_ep.signature_change	|Yes	|No	|No	|
|SCTP_TIMER_TYPE_PATHMTURAISE	|net->pmtu_timer		|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_SHUTDOWNACK	|net->rxt_timer			|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_ASCONF		|stcb->asoc.asconf_timer	|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_SHUTDOWNGUARD	|stcb->asoc.shut_guard_timer	|Yes	|Yes	|No	|
|SCTP_TIMER_TYPE_AUTOCLOSE	|stcb->asoc.autoclose_timer	|Yes	|Yes	|No	|
|SCTP_TIMER_TYPE_STRRESET	|stcb->asoc.strreset_timer	|Yes	|Yes	|Yes	|
|SCTP_TIMER_TYPE_INPKILL	|inp->sctp_ep.signature_change	|Yes	|No	|No	|
|SCTP_TIMER_TYPE_ASOCKILL	|stcb->asoc.strreset_timer	|Yes	|Yes	|No	|
|SCTP_TIMER_TYPE_ADDR_WQ	|SCTP_BASE_INFO(addr_wq_timer)	|No	|No	|No	|
|SCTP_TIMER_TYPE_PRIM_DELETED	|stcb->asoc.delete_prim_timer	|Yes	|Yes	|No	|

taylor-b:
@tuexen That's correct. It follows the pattern of "if you reference something but don't hold its lock, you need to hold a reference," which is what I saw the rest of the codebase doing.

mstyura:
Hello @tuexen and @taylor-b,
Hope you and your families feel well and safe!
I've got slightly lost on the status of this PR. I've seen many parts from PR was kind of "cherry picked" into master already, but some of the changes are still not merged. Are they pending further work or abandoned due to not being necessary anymore? I see that Chromium switched to some commit which is no longer directly reachable from `master` probably due to force push happen... @taylor-b maybe it make sense to adjust commit in `DEPS` in `chromium` and `webrtc` as well?

Thanks a lot in advance!

tuexen:
@mstyura I'm in the process of getting it in. The current state is that I cleaned up the timer code to avoid cases where we would increment/decrement in a way we shouldn't. We now have KASSERTs in place and some fuzz tester are finding issues (syzkaller based). Once these bugs are  fixed, adding the refcounts is a simple change. That is what I'm working towards.

mstyura:
Thank you very much @tuexen!

Lastique:
If this PR is still relevant, could it be rebased on the current master please?

tuexen:
> If this PR is still relevant, could it be rebased on the current master please?

It is still relevant. I'm still cleaning up the counters before making use of them in this context.

tuexen:
A modified version of this PR has been committed in https://github.com/sctplab/usrsctp/commit/939d48f9632d69bf170c7a84514b312b6b42257d.

taylor-b:
In that case, closing.

PR #433, opened 2020-02-17
usrSCTP no longer emits invalid fragmented packets with prio SS

### LINKED ISSUE #432: SCTP stream blocks up intermittently with large packets
Our use case is to use SCTP to send large frames as well as lots of very small frames. The large and small frames are on different SCTP streams to avoid head-of-line blocking. However, we have observed that occasionally the stream with which we are sending large frames on blocks up, and we no longer receive any more data.

This seems to happen when at the receiver a packet arrives on one stream in the middle of a fragmented packet arriving on another. Once that happens, we never again see callbacks[1] to the upper layer from usrSCTP on that particular stream.

I have attached two files from the receiving usrSCTP stack:
* The first is a log of the usrSCTP output with SCTP_DEBUG defined
* The second is a wireshark capture on the interface where the packets arrive

Note that we have a custom header in between the UDP header and SCTP header which is 6 bytes big. The first two bytes of these is a packet type in LE, a value of 0 indicates a data packet to be passed to usrSCTP, anything else is not passed down to usrSCTP and can be ignored. I have attached a dissector for wireshark that will do this, set a filter of "cw_header.type == 0" to ignore the irrelevant packets.

The point where we see the failure is on packet 44384 of the capture, which is received at line 373715 of the log file. There are a couple of received fragmented packets, then the off-stream packet arrives (packet 44393, line 373787) and when the last fragment arrives (packet 44409, line 373817), we never get a callback on the high-data stream again.

I will be investigating this deeper next week to try and resolve the issue, but any thoughts or ideas would be very helpful!

The files mentioned above are uploaded in the following zip file.
[usrsctp_issues_files.zip](https://github.com/sctplab/usrsctp/files/4202328/usrsctp_issues_files.zip)

[1] We're using the callback API rather than recvv, to avoid running a thread per socket as we're opening a couple of sockets

somtin:
This resolves #432

### DIFF
--- a/usrsctplib/netinet/sctp_ss_functions.c
+++ b/usrsctplib/netinet/sctp_ss_functions.c
@@ -521,6 +521,9 @@ sctp_ss_prio_select(struct sctp_tcb *stcb SCTP_UNUSED, struct sctp_nets *net,
 {
 	struct sctp_stream_out *strq, *strqt, *strqn;
 
+	if (asoc->ss_data.locked_on_sending) {
+		return (asoc->ss_data.locked_on_sending);
+	}
 	strqt = asoc->ss_data.last_out_stream;
 prio_again:
 	/* Find the next stream to use */


tuexen:
Good catch. Let me double check if other schedulers are also affected.

tuexen:
Two other schedulers are also affected. I checked in a fix in https://github.com/sctplab/usrsctp/commit/dc83f15ee6c7c54bb77580d88a96029c84dd43b3.
Thanks for reporting and providing a fix.

PR #438, opened 2020-03-06
Fix error with -fsanitize=alignment.

### LINKED ISSUE #437: Error with -fsanitize=alignment
Related WebRTC bug: https://bugs.chromium.org/p/webrtc/issues/detail?id=11405

The issue is with sctp_getaddresses, when casting addr to a sockaddr_conn.

On a 64 bit system, sockaddr_conn is 8-byte aligned due to one of its members being a pointer. However, sockaddr is only 4-byte aligned, so the compiler doesn't know it needs to insert four bytes of padding.

Fix on the way.

taylor-b:
The issue is with sctp_getaddresses, when casting addr to a sockaddr_conn.

On a 64 bit system, sockaddr_conn is 8-byte aligned due to one of its members
being a pointer. However, sockaddr is only 4-byte aligned, so the compiler
doesn't know it needs to insert four bytes of padding.

The solution is simply to use a union of sockaddr and sockaddr_conn.

Fixes #437.

### DIFF
--- a/usrsctplib/netinet/sctp_uio.h
+++ b/usrsctplib/netinet/sctp_uio.h
@@ -662,23 +662,19 @@ struct sctp_setpeerprim {
 	uint8_t sspp_padding[4];
 };
 
-/*
- * This is necessary to prevent misaligned access to a sockaddr_conn.
- *
- * On a 64 bit system, sockaddr_conn is 8-byte aligned due to one of its
- * members being a pointer. However, sockaddr is only 4-byte aligned, so
- * without this union the compiler would not insert the necessary padding
- * between sget_assoc_id and addr.
- */
-union sockaddr_union {
-	struct sockaddr addr;
-	struct sockaddr_conn addr_conn;
+union sctp_sockstore {
+	struct sockaddr_in sin;
+	struct sockaddr_in6 sin6;
+#if defined(__Userspace__)
+	struct sockaddr_conn sconn;
+#endif
+	struct sockaddr sa;
 };
 
 struct sctp_getaddresses {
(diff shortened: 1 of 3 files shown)


alvestrand:
Pinging @tuexen for review. Please see associated https://crbug.com/webrtc/11405 for more comments.

tuexen:
@alvestrand Providing links like https://crbug.com/webrtc/11405 doesn't really help. I also get once in a while mails with them. However, when trying to access the bug reports I always get 'access denied'. So I'm ignoring these mails... So if you want to provide some information, put it in an email.

I'll look into this issue.

Lastique:
Building this PR with c5abb7a4e65449e32c33fa7703a7bc234bdffb82 fails:

```
user_socket.c: In function ‘usrsctp_bindx’:
user_socket.c:2743:24: error: ‘union sockaddr_union’ has no member named ‘sa_family’
 2743 |    switch (gaddrs->addr->sa_family) {
      |                        ^~
```

Lastique:
Maybe a better fix would be to change `sctp_getaddresses::addr` type to array of `sockaddr_storage`, which should already have the proper alignment.

taylor-b:
`sockaddr_storage` doesn't include `sockaddr_conn` though, which was the problem here. Though I noticed there is already a `sctp_sockaddr` union defined in this header so I'm just using that.

Also fixed compilation errors.

Lastique:
Thanks, the updated PR compiles without errors.

taylor-b:
Fixed merge conflicts.

tuexen:
This should be fixed now.

PR #448, opened 2020-03-20
Fix gcc compilation error

lminiero:
The commit d07a5f27630b1250655337791adf2a4c171a6712 from yesterday causes gcc to fail with this error in `sctp_usrreq.c`:

```
netinet/sctp_usrreq.c: In function ‘sctp_disconnect’:
netinet/sctp_usrreq.c:1279:23: error: variable ‘netp’ set but not used [-Werror=unused-but-set-variable]
 1279 |     struct sctp_nets *netp;
      |                       ^~~~
```

Apparently it doesn't trigger any warning when using clang instead. This pull request simply gets rid of the `netp` inizialitation, since where the error occurs the variable is not used anymore (`NULL` is now always passed to `sctp_timer_start`).

### DIFF
--- a/usrsctplib/netinet/sctp_usrreq.c
+++ b/usrsctplib/netinet/sctp_usrreq.c
@@ -1276,13 +1276,6 @@ sctp_disconnect(struct socket *so)
 				 * we will allow user data to be sent first
 				 * and move to SHUTDOWN-PENDING
 				 */
-				struct sctp_nets *netp;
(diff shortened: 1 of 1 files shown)


tuexen:
I committed a similar fix which also include the FreeBSD line (after committing it there). Thanks for the report!

lminiero:
Thanks for the quick fix!

PR #450, opened 2020-03-28
Fix OOB read with sctp_getopt(SCTP_STATUS) and AF_CONN

taylor-b:
The code was assuming the address was either AF_INET or AF_INET6, resulting in an out of bounds read if it's actually AF_CONN.

I'm pretty confident this will just be reading some extra zeros, since new `sctp_nets`s are always memset to 0. But still, might as well fix it to err on the safe side.

### DIFF
--- a/usrsctplib/netinet/sctp_usrreq.c
+++ b/usrsctplib/netinet/sctp_usrreq.c
@@ -3390,10 +3390,17 @@ sctp_getopt(struct socket *so, int optname, void *optval, size_t *optsize,
 				memcpy(&sstat->sstat_primary.spinfo_address,
 				       &stcb->asoc.primary_destination->ro._l_addr,
 				       sizeof(struct sockaddr_in));
-			} else {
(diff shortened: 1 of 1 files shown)


tuexen:
If we would reach the else error path, it is not a user error, but something very strange, because the address family of the primary path would be one we don't support. So panic()ing would make more sense than indicating a user that there is a problem.

leaving here with break would leave the stcb lock locked.

I committed a fix, which uses switch/case, in https://github.com/sctplab/usrsctp/commit/0390f325a39b067d2177b7fc4f2a08ae94a0ac6d.

Thanks for reporting the issue!

taylor-b:
Thanks!

Manouchehri:
My bad for making the incorrect break statement, wasn't super familiar with the code. Originally reported over here if you'd like more context: https://bugs.chromium.org/p/chromium/issues/detail?id=1065253

Issue #299, opened 2019-02-15
Do no undefine __APPLE__ on mac/ios builds


pps83rbx:
This was already reported and [simply closed before](https://github.com/sctplab/usrsctp/issues/152). This is a bug that has to be fixed.

`__APPLE__` is a known identifier for apple builds (mac/ios), this is insane to undefine it. More over, this isn't supported, as identifiers with leading underscores are reserved. You aren't supposed to define/undefine these, I hope it's not a surprise for you. If you need something special, then it should have been SCTP_APPLE or whatever.
The way it is now is kind of crazy: some structs have different layout depending of presence of `__APPLE__` define. I hope you do not expect that any unrelated code that happens to use usrsctp library will also need to undefine `__APPLE__` in their builds!? I also hope that tests for `__APPLE__` in public headers doesn't affect layouts of these strucs in way that would corrupt anything that tries to use usrsctp on mac/ios. I'd be scared to touch this library with a long pole in a commercial product if I didn't do full code review for each use of `__APPLE__`. Please, fix the bug to avoid these issues.

lgrahl:
@tuexen already said he agrees with you and that they are very much willing to accept a pull request.

pps83rbx:
I just did quick code review and I'm totally shocked that this code doesn't bomb. It' sued by chrome in webrtc, strange that they didn't notice it.

lgrahl:
I also did a quick review... this is dead code unless someone defines `SCTP_USE_NSS_SHA1`. So, you can calm down now.

lgrahl:
And apparently none of the browsers define it, so they aren't aware (or were aware and decided not to report it):

* Chrome: https://cs.chromium.org/chromium/src/third_party/usrsctp/BUILD.gn?l=96&rcl=38de9b980a6fd8ca452c39b2db1dec5639459976
* Firefox: https://dxr.mozilla.org/mozilla-central/search?q=SCTP_USE_NSS_SHA1&redirect=false
* WebKit: https://github.com/WebKit/webkit/blob/9d678322281ccd07909759af90ee010e0c34b1c7/Source/ThirdParty/libwebrtc/CMakeLists.txt#L1499

pps83rbx:
> I also did a quick review... this is dead code unless someone defines `SCTP_USE_NSS_SHA1`. So, you can calm down now. 🙂

I wasn't at all concerned by this piece. It's "safe" even though ugly hasck as the code says.

These are the problems I'm concerned about:

[...]

PR #451, opened 2020-04-03
Clang module compatibility

jgh-:
I've been working on using usrsctp with a swift package and found some incompatibilities with the swift package manager and clang modules:

1. Undeffing `__APPLE__` is not a workable solution because it requires using unsafe compiler flags and therefore breaks the swift package (unsafe swift packages cannot be used as dependencies).  So I've renamed the OS checks which get undeffed to `SCTP_KERNEL_xxx` which now need to be activated if those features are desired, rather than fighting the compiler toolchain by undeffing default definitions. So this addresses #299 

2. Renamed `mbstat`, `xsocket` and `xsockbuf` to `usr_mbstat`, `usr_xsocket` and `usr_xsockbuf` so that they don't conflict with the system definitions of those structs.  This isn't a problem when you aren't using modules, but once you enable modules the build breaks because of this conflict.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -46,11 +46,11 @@ APPCFLAGS=""
 case $host_os in
 darwin*)
   CFLAGS="$CFLAGS -std=c99 -Wno-deprecated-declarations -D__APPLE_USE_RFC_2292"
-  LIBCFLAGS="$LIBCFLAGS -U__APPLE__ -D__Userspace_os_Darwin"
+  LIBCFLAGS="$LIBCFLAGS -D__Userspace_os_Darwin"
   ;;
 dragonfly*)
   CFLAGS="$CFLAGS -std=c99 -pthread"
-  LIBCFLAGS="$LIBCFLAGS -U__DragonFly__ -D__Userspace_os_DragonFly"
+  LIBCFLAGS="$LIBCFLAGS -D__Userspace_os_DragonFly"
   ;;
 freebsd*)
   CFLAGS="$CFLAGS -std=c99 -pthread"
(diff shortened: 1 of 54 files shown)


tuexen:
I remove `mbstat`, `xsocket`, and `xsockbuf`  in https://github.com/sctplab/usrsctp/commit/7f8671560bd7856583b81d17913515249067607d, since they are not really used.

Please note that the defines used in directories other than `netinet` and `netinet6` should be fine. So changing them in `programs`, for example, breaks the programs (they are not running in the kernel...).

I have started cleaning up defines. Support for `__Panda__` is gone. The next step will be only to support the FreeBSD kernel in head, not older versions of it. This was used in the past, but is not used anymore.

Then the plan is to do some replacements:

|Old | New|
|----|-----|
|`defined(__APPLE__)`|`defined(__APPLE__) && defined(_KERNEL)`|
|`defined(__FreeBSD__)`|`defined(__FreeBSD__) && defined(_KERNEL)`|
|`defined(__Userspace__)`|`!defined(_KERNEL)`|
|`defined(__Userspace_os_Darwin)`|`defined(__APPLE__) && !defined(_KERNEL)`|
|`defined(__Userspace_os_FreeBSD)`|`defined(__FreeBSD__) && !defined(_KERNEL)`|

I'm not sure about the Windows related ones... The reason to do the cleanup first is to simplify the conditions before changing them...

tuexen:
The problem with `-U__APPLE` should be fixed by https://github.com/sctplab/usrsctp/commit/5be0c259659dd3f810011a52676b59a0e4706b0a. If not, please open an issue.

PR #461, opened 2020-05-01
user_recv_thread: set fd to -1 before closing

ystreet:
Fixes a busy loop of the usrsctp UDP recv thread attempting to shutdown
usrsctp with usrsctp_finish().

Usually, destroying the udp socket in recv_thread_destroy() would result
in the udp receive thread returning an error (-1) with errno set to
(-9) EDABF.  In other cases, sometimes recvmsg would return 0 without an
error being reported in errno.  My assumption here is that the fd number
would be reused for some other purpose.

Explicitly set the socket/fd variables to -1 before performing a
close()/closesocket() on them so that any future recv*() calls will
return with an error.

### DIFF
--- a/usrsctplib/netinet/sctp_usrreq.c
+++ b/usrsctplib/netinet/sctp_usrreq.c
@@ -211,49 +211,6 @@ sctp_finish(void)
 #if defined(INET) || defined(INET6)
 	recv_thread_destroy();
 #endif
-#if !defined(__Userspace_os_Windows)
-#if defined(INET) || defined(INET6)
-	if (SCTP_BASE_VAR(userspace_route) != -1) {
-		pthread_join(SCTP_BASE_VAR(recvthreadroute), NULL);
-	}
-#endif
-#endif
-#ifdef INET
-	if (SCTP_BASE_VAR(userspace_rawsctp) != -1) {
-#if defined(__Userspace_os_Windows)
-		WaitForSingleObject(SCTP_BASE_VAR(recvthreadraw), INFINITE);
-		CloseHandle(SCTP_BASE_VAR(recvthreadraw));
-#else
-		pthread_join(SCTP_BASE_VAR(recvthreadraw), NULL);
-#endif
-	}
-	if (SCTP_BASE_VAR(userspace_udpsctp) != -1) {
-#if defined(__Userspace_os_Windows)
-		WaitForSingleObject(SCTP_BASE_VAR(recvthreadudp), INFINITE);
-		CloseHandle(SCTP_BASE_VAR(recvthreadudp));
-#else
-		pthread_join(SCTP_BASE_VAR(recvthreadudp), NULL);
-#endif
-	}
-#endif
-#ifdef INET6
-	if (SCTP_BASE_VAR(userspace_rawsctp6) != -1) {
-#if defined(__Userspace_os_Windows)
-		WaitForSingleObject(SCTP_BASE_VAR(recvthreadraw6), INFINITE);
-		CloseHandle(SCTP_BASE_VAR(recvthreadraw6));
-#else
-		pthread_join(SCTP_BASE_VAR(recvthreadraw6), NULL);
-#endif
-	}
-	if (SCTP_BASE_VAR(userspace_udpsctp6) != -1) {
-#if defined(__Userspace_os_Windows)
-		WaitForSingleObject(SCTP_BASE_VAR(recvthreadudp6), INFINITE);
-		CloseHandle(SCTP_BASE_VAR(recvthreadudp6));
-#else
-		pthread_join(SCTP_BASE_VAR(recvthreadudp6), NULL);
-#endif
-	}
-#endif
 	atomic_cmpset_int(&SCTP_BASE_VAR(timer_thread_should_exit), 0, 1);
 	if (SCTP_BASE_VAR(timer_thread_started)) {
 #if defined(__Userspace_os_Windows)
--- a/usrsctplib/user_recv_thread.c
+++ b/usrsctplib/user_recv_thread.c
@@ -1456,6 +1456,7 @@ recv_thread_destroy(void)
 #if defined(INET) || defined(INET6)
 	if (SCTP_BASE_VAR(userspace_route) != -1) {
 		close(SCTP_BASE_VAR(userspace_route));
+		pthread_join(SCTP_BASE_VAR(recvthreadroute), NULL);
 	}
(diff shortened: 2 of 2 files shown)


tuexen:
> Fixes a busy loop of the usrsctp UDP recv thread attempting to shutdown
> usrsctp with usrsctp_finish().
> 
> Usually, destroying the udp socket in recv_thread_destroy() would result
> in the udp receive thread returning an error (-1) with errno set to
> (-9) EDABF. In other cases, sometimes recvmsg would return 0 without an
> error being reported in errno.
> 
> From the POSIX man page, the recv*() family of functions can return 0
> indicating no messages and an orderly shutdown:
> 
> "If no messages are available to be received and the peer has performed
> an orderly shutdown, recv() shall return 0"

This only applies to connection oriented sockets (like for TCP and SCTP), not for UDP sockets. It is perfectly fine to receive a UDP packet without user data and recv() returns in that case 0.

ystreet:
> This only applies to connection oriented sockets (like for TCP and SCTP), not for UDP sockets. It is perfectly fine to receive a UDP packet without user data and recv() returns in that case 0. UDP does not support a connection setup or shutdown.

Hrm, so it is.  In that case, updated with an adjusted patch that is more careful with fd closure and the variables involved.

tuexen:
I'm still not sure what you want to achieve... Right now you can't set the `fd`s to `-1`, since this is used to indicate if a corresponding thread was started. See `sctp_finish()`. I assume this is called at the end of a program, where I would not expect that other parts of the program open sockets again.

So what is the problem you are experiencing?

ystreet:
The problem is that stopping usrsctp can busy loop in the receive threads while sctp_finish() is trying to join those threads.  Both of the `recv_function_udp` and `recv_function_udp6` are busy looping in this scenario due to what I believe to be fd reuse.  Here is a backtrace of the relevant usrsctp threads.

```
Thread 273 (Thread 0x7fff5d7f2700 (LWP 470887)):
#0  0x00007ffff7b542d1 in clock_nanosleep@GLIBC_2.2.5 () at /usr/lib/libc.so.6
#1  0x00007ffff7b59bf7 in nanosleep () at /usr/lib/libc.so.6
#2  0x00007ffff504b63d in user_sctp_timer_iterate (arg=<optimized out>) at ../subprojects/usrsctp/usrsctplib/netinet/sctp_callout.c:209
#3  0x00007ffff786a46f in start_thread () at /usr/lib/libpthread.so.0
#4  0x00007ffff7b8c3d3 in clone () at /usr/lib/libc.so.6

Thread 272 (Thread 0x7fff5dff3700 (LWP 470886)):
#0  0x00007ffff7b8d75d in recvmsg () at /usr/lib/libc.so.6
#1  0x00007ffff503c69b in recv_function_udp6 (arg=<optimized out>) at ../subprojects/usrsctp/usrsctplib/user_recv_thread.c:903
#2  0x00007ffff786a46f in start_thread () at /usr/lib/libpthread.so.0
#3  0x00007ffff7b8c3d3 in clone () at /usr/lib/libc.so.6

Thread 271 (Thread 0x7fff5e7f4700 (LWP 470885)):
#0  0x00007ffff7b8d75d in recvmsg () at /usr/lib/libc.so.6
#1  0x00007ffff503d08e in recv_function_udp (arg=<optimized out>) at ../subprojects/usrsctp/usrsctplib/user_recv_thread.c:701
#2  0x00007ffff786a46f in start_thread () at /usr/lib/libpthread.so.0
#3  0x00007ffff7b8c3d3 in clone () at /usr/lib/libc.so.6

Thread 270 (Thread 0x7fff5eff5700 (LWP 470884)):
#0  0x00007ffff7870cf5 in pthread_cond_wait@@GLIBC_2.3.2 () at /usr/lib/libpthread.so.0
#1  0x00007ffff504b023 in sctp_iterator_thread (v=<optimized out>) at ../subprojects/usrsctp/usrsctplib/netinet/sctp_bsd_addr.c:151
#2  0x00007ffff786a46f in start_thread () at /usr/lib/libpthread.so.0
#3  0x00007ffff7b8c3d3 in clone () at /usr/lib/libc.so.6
...
Thread 1 (Thread 0x7ffff7416740 (LWP 470611)):
#0  0x00007ffff786ba67 in __pthread_clockjoin_ex () at /usr/lib/libpthread.so.0
#1  0x00007ffff5089b12 in sctp_finish () at ../subprojects/usrsctp/usrsctplib/netinet/sctp_usrreq.c:235
#2  0x00007ffff50407c3 in usrsctp_finish () at ../subprojects/usrsctp/usrsctplib/user_socket.c:2249
#3  usrsctp_finish () at ../subprojects/usrsctp/usrsctplib/user_socket.c:2235
#4  0x00007ffff61bd495 in gst_sctp_association_finalize (object=0x7fffc0002ec0) at ../subprojects/gst-plugins-bad/ext/sctp/sctpassociation.c:245
...
```

We have no guarantee that other threads are not opening any sockets or fd's at this time and cannot provide that guarantee.

tuexen:
The only solution I see is to add a boolean for each thread which tells us if the corresponding thread was started and we need to wait for joining it. Then we can just close the sockets and set the fd to -1.

ystreet:
I thought about that, but then decided to move the thread joining from `sctp_finish()` to the `recv_thread_destroy()` instead as that's the only use of `recv_thread_destroy()`.  I think that achieves similar results as an extra boolean for each thread.  This code (both before and after this change) is relying on the `close()` to be atomic against the `recvmsg()`.  If that is ever ot the case, then I think a cancellable fd would need to be added that can be `select()/poll()`ed on with the original fd.

Calling `pthread_join()` directly after `close()` introduces a noticeable wait (100-200ms per thread here) through and it may be more prudent to close all the fd's and then join all of the threads to introduce a little more parallelism.  If you want to do that, I can implement that as well.

tuexen:
I guess the 100ms delay comes from the socket timeout, which is used. Since you should use the stack destruction only when you teardown your program, I don't think we need to optimise here.

I don't think that you gain much by using
```
fd = SCTP_BASE_VAR(var);
SCTP_BASE_VAR(var) = -1;
close(fd);
pthread_join();
```
instead of
```
close(SCTP_BASE_VAR(var));
SCTP_BASE_VAR(var) = -1;
pthread_join();
```
Do you observe a difference?

ystreet:
I don't think there is a noticeable difference.

tuexen:
Merged. Thanks for the contribution.

PR #467, opened 2020-05-14
Add usrsctp_get_ulpinfo getter.

### LINKED ISSUE #465: Feature request: usrsctp_get_ulpinfo
When creating a socket, you pass in a `void* ulp_info` which is also passed into the receive callback for that socket, which can conveniently be used to associate the socket with a C++ class or whatnot.

However, that pointer is not passed into the send callback. And although there is a `usrsctp_set_ulpinfo` function, there is no equivalent getter.

Is there any reason not to add one? Due to its absence, we've been using the `sockaddr_conn` address to store a pointer value, but it's come to our attention that [that's not a good idea](https://bugs.chromium.org/p/chromium/issues/detail?id=1076703).

taylor-b:
Can be used to access `ulp_info` from within `send_cb`.

Fixes #465.


tuexen:
Thanks for the patch. Perfect!

PR #470, opened 2020-05-19
Adjust `usrsctp_handle_timers` signature and implementation

mstyura:
Currently `usrsctp_handle_timers` passes delta value came from user application as is into `sctp_handle_tick`.
This only work correctly when delta value is measured in ticks.
As long as there is no way (or I can't find one) to figure out how ticks translated to wall clock time there is no way to use of this method pass correct ticks value.
I believe it is correct to allow user to pass elapsed milliseconds, which is converted by library to internal notion of time - ticks.

### DIFF
--- a/programs/st_client.c
+++ b/programs/st_client.c
@@ -68,9 +68,10 @@ static int connecting = 0;
 static int finish = 0;
 
 static unsigned int
-get_tick_count(void)
+get_milliseconds_count(void)
 {
 #ifdef _WIN32
+	// obtain number of milliseconds since system started
 	return GetTickCount();
 #else
 	struct timeval tv;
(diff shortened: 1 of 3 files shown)


mstyura:
Hello @tuexen!
I hope you and your family are well and safe.
Could you please have a look at this PR, which I believe fixes an issue with "single threaded API", when timer handler is invoked by user application?

Thanks a lot in advance!

tuexen:
That is a change of the API, but it really makes sense. So if no-one objects, I'm inclined to merge.

mstyura:
@tuexen I agree that there is only potential problem, because by default `HZ=1000`. It also mean that for an existing users this change is very likely to be no-op, while making things more explicit. Since nobody objected so far maybe it worth merging this PR? Thanks a lot!

PR #339, opened 2019-08-07
Single thread

jmillan:
Based on https://github.com/sctplab/usrsctp/pull/38/commits/010fcbc6f193152607dc267c7e7e573c9bffc111

PR #472, opened 2020-05-22
Do not start `SCTP iterator` when initialized via `usrsctp_init_nothreads`.

mstyura:
I believe it make sense not to stat `SCTP iterator` thread, when library was intentionally initialized via `usrsctp_init_nothreads`.

In case user of library desire to disable threads there should no threads created.

I see main use cases of `usrsctp_init_nothreads` is to implement single threaded processing of `SCTP` packets for applications which need to implement WebRTC's data channel.

It is occurred that completely single threaded processing is impossible due to `SCTP iterator` is still started when library initialized via `usrsctp_init_nothreads`. 

Moreover this thread is still doing some work - it is not always in dormant state.
So far I've figured out that some processing on `SCTP iterator` is triggered by timer, with example call stack:
```
sctp_wakeup_iterator() usrsctplib/netinet/sctp_bsd_addr.c:114)
sctp_initiate_iterator(inp_func inpf, asoc_func af, inp_func inpe, uint32_t pcb_state, uint32_t pcb_features, uint32_t asoc_state, void * argp, uint32_t argi, end_func ef, struct sctp_inpcb * s_inp, uint8_t chunk_output_off) usrsctplib/netinet/sctp_pcb.c:8207)
sctp_handle_addr_wq() usrsctplib/netinet/sctputil.c:1725)
sctp_timeout_handler(void * t) usrsctplib/netinet/sctputil.c:2175)
sctp_handle_tick(uint32_t elapsed_ticks) usrsctplib/netinet/sctp_callout.c:172)
usrsctp_handle_timers(uint32_t delta) usrsctplib/user_socket.c:3548)
```
https://github.com/sctplab/usrsctp/blob/79ce3f1e13cd1cb283871ec5a9f90cf4062d91d4/usrsctplib/netinet/sctputil.c#L2171-L2176
https://github.com/sctplab/usrsctp/blob/79ce3f1e13cd1cb283871ec5a9f90cf4062d91d4/usrsctplib/netinet/sctputil.c#L1694-L1746
https://github.com/sctplab/usrsctp/blob/79ce3f1e13cd1cb283871ec5a9f90cf4062d91d4/usrsctplib/netinet/sctp_pcb.c#L8121-L8207

@tuexen do you think it is possible to accept this PR to make library behave as promised by name of `usrsctp_init_nothreads`? Thanks a lot in advance!

### DIFF
--- a/usrsctplib/netinet/sctp_bsd_addr.c
+++ b/usrsctplib/netinet/sctp_bsd_addr.c
@@ -108,6 +108,11 @@ sctp_cleanup_itqueue(void)
 void
 sctp_wakeup_iterator(void)
 {
+	if (!SCTP_BASE_VAR(iterator_thread_started)) {
+		// no dedicated thread stated, so execute
+		// iterator work on current thread
+		return sctp_iterator_worker();
+	}
 #if defined(SCTP_PROCESS_LEVEL_LOCKS)
 #if defined(__Userspace_os_Windows)
 	WakeAllConditionVariable(&sctp_it_ctl.iterator_wakeup);
@@ -177,7 +182,11 @@ sctp_iterator_thread(void *v SCTP_UNUSED)
 }
 
 void
+#if defined(__Userspace__)
+sctp_startup_iterator(int start_threads)
+#else
 sctp_startup_iterator(void)
+#endif
 {
 	if (sctp_it_ctl.thread_proc) {
 		/* You only get one */
@@ -187,6 +196,13 @@ sctp_startup_iterator(void)
 	SCTP_ITERATOR_LOCK_INIT();
 	SCTP_IPI_ITERATOR_WQ_INIT();
 	TAILQ_INIT(&sctp_it_ctl.iteratorhead);
+
+#if defined(__Userspace__)
+	if (!start_threads) {
+		return;
+	}
+#endif
+
 #if defined(__FreeBSD__)
 #if __FreeBSD_version <= 701000
 	kthread_create(sctp_iterator_thread,
--- a/usrsctplib/netinet/sctp_bsd_addr.h
+++ b/usrsctplib/netinet/sctp_bsd_addr.h
@@ -47,7 +47,12 @@ __FBSDID("$FreeBSD: head/sys/netinet/sctp_bsd_addr.h 353480 2019-10-13 18:17:08Z
 extern struct iterator_control sctp_it_ctl;
 void sctp_wakeup_iterator(void);
 
-void sctp_startup_iterator(void);
+void
+#if defined(__Userspace__)
+sctp_startup_iterator(int start_threads);
+#else
+sctp_startup_iterator(void);
+#endif
 
 
 #ifdef INET6
--- a/usrsctplib/netinet/sctp_pcb.c
+++ b/usrsctplib/netinet/sctp_pcb.c
@@ -6799,7 +6799,12 @@ sctp_pcb_init(void)
 	(void)pthread_cond_init(&sctp_it_ctl.iterator_wakeup, NULL);
 #endif
 #endif
+
+#if defined(__Userspace__)
+	sctp_startup_iterator(start_threads);
+#else
 	sctp_startup_iterator();
+#endif
 
 #if defined(__FreeBSD__) && defined(SCTP_MCORE_INPUT) && defined(SMP)
 	sctp_startup_mcore_threads();


mstyura:
Dr. @tuexen, could you please have a look at this PR and share your thought if these changes are something that makes sense to accept? 
PS: original PR which added single thread API is https://github.com/sctplab/usrsctp/pull/339

tuexen:
The problem I see with  this approach is that the  `sctp_wakeup_iterator()` might take a long time now. The idea of the iterator is to delegate work items which might take a long time. With your proposed patch this is not true anymore.

mstyura:
Thanks you very much for your time and viewing this PR Dr. @tuexen!

Yes, that also my concern. But when I proposed this change I thought of following:
1. Primary use case for single threaded API from my point of view are sockets created with `AF_CONN` - so basically `usrsctp` is used to prepare `sctp` packets from user data to being sent by other transport (e.g. `DTLS`) and assemble data from `sctp` packets, which are provided by user application. This is solely CPU-bound work in my understanding.
2. Taking into account the nature of `AF_CONN` sockets (as far as I understand them) there is by definition no I/O or blocking operations (other than `mutex` acquisition, which should not be a problem when there is only one thread) is done inside library. Hence any work which could potentially done by `SCTP Iterator` is `CPU`-bound and hence it could be done in the same thread as other operations, when library is initialized in the single threaded API.
3. Even if there is some potentially long running `CPU`-bound tasks done by `SCTP iterator` thread then it would be expected to be done on user thread, when library is inited via `usrsctp_init_nothreads`. 

Maybe my understanding of `AF_CONN` sockets is incorrect and there are still I/O or blocking operations done by library, if so then yes, there is a reason for a dedicated thread, but it still unclear if it is expected behavior when inited via `usrsctp_init_nothreads`... Maybe signature of method could be changed to accept `2` bools: `1` for `SCTP timer` thread and `1` for `SCTP iterator` thread?

Do you think it is possible to change signature of `usrsctp_init_nothreads` to not to start `SCTP timer` and have a "true" single threaded API mode?

Thank you very much!

tuexen:
> Thanks you very much for your time and viewing this PR Dr. @tuexen!
> 
> Yes, that also my concern. But when I proposed this change I thought of following:
> 
> 1. Primary use case for single threaded API from my point of view are sockets created with `AF_CONN` - so basically `usrsctp` is used to prepare `sctp` packets from user data to being sent by other transport (e.g. `DTLS`) and assemble data from `sctp` packets, which are provided by user application. This is solely CPU-bound work in my understanding.

I was also assuming for some time that you would use the userland stack in the context of WebRTC (with AF_CONN sockets). For native SCTP you would use a kernel stack. But people are using, as far as I can tell from bug reports, it also for native SCTP usage.

> 2. Taking into account the nature of `AF_CONN` sockets (as far as I understand them) there is by definition no I/O or blocking operations (other than `mutex` acquisition, which should not be a problem when there is only one thread) is done inside library. Hence any work which could potentially done by `SCTP Iterator` is `CPU`-bound and hence it could be done in the same thread as other operations, when library is initialized in the single threaded API.

The question is: When you are using the library in single threaded mode, are you expecting calls to the library taking an arbitrary long time?

> 3. Even if there is some potentially long running `CPU`-bound tasks done by `SCTP iterator` thread then it would be expected to be done on user thread, when library is inited via `usrsctp_init_nothreads`.
> 
> Maybe my understanding of `AF_CONN` sockets is incorrect and there are still I/O or blocking operations done by library, if so then yes, there is a reason for a dedicated thread, but it still unclear if it is expected behavior when inited via `usrsctp_init_nothreads`... Maybe signature of method could be changed to accept `2` bools: `1` for `SCTP timer` thread and `1` for `SCTP iterator` thread?

Again, one has to provide a way how the library is expected to be used.

> 
> Do you think it is possible to change signature of `usrsctp_init_nothreads` to not to start `SCTP timer` and have a "true" single threaded API mode?

Maybe a possible way is to expose a `usrsctp_iterator()` function, which you have to call periodically and which can run arbitrarily long. Some stuff (like using a recently added address) might only work after calling the `usrsctp_iterator()`. Would that work for you?

> 
> Thank you very much!

ibc:
IMHO, definitely a library that is supposed to run in single thread mode (or that provides an option for it, such as `usrsctp_init_nothreads`) should not run other threads internally, and whatever such a thread does should be exposed to the application so it can provide similar behavior without requiring such a thread.

tuexen:
Please note the `usrsctp_init_nothreads()` was a contribution. I assumed it was useful for the contributor.

ibc:
> Please note the `usrsctp_init_nothreads()` was a contribution. I assumed it was useful for the contributor.

:)

PR #474, opened 2020-05-25
Fixed compilation due usage of non existing macro

mstyura:
Not sure how that happen, but in #470 I've used macro which no longer exists and it went unnoticed by me.
So, this PR is fixing the issue.


mstyura:
Dr. @tuexen,
sorry for bothering you again, it is occurred that by accident my previous PR introduced compilation error. And this PR fixes an issue.
Thanks a lot!

tuexen:
No problem.

PR #481, opened 2020-06-11
Adding some CORPUS for packet-dropped chunks

yuquanw2:


tuexen:
Any reason you want to test that? The corresponding ID has expired and we don't enable this by default. This is not an argument against merging, I'm just making sure that there are not wrong expectations in relation to this extension.

yuquanw2:
Thank you for commenting! Just as you said, I am aware that this code is generally disabled. And no plans to use it either. For now, I am just working on adding fuzzing coverage for every feature in case there might be some usage in the future.

tuexen:
OK, thanks a lot!

PR #485, opened 2020-06-19
tsctp: add partial reliability with parameters -P and -t

msvoelker:
Adds parameters to the tsctp program in order to use partial reliability (-P and -t). To set the PR values, an sctp_sendv_spa struct is used for usrsctp_sendv instead of an sctp_sndinfo struct.


programs/tsctp.c
```diff
@@ -762,14 +816,45 @@ int main(int argc, char **argv)
...
+				sendv_spa.sendv_sndinfo.snd_flags |= SCTP_UNORDERED;
+			}
+			sendv_spa.sendv_sndinfo.snd_ppid = 0;
+			sendv_spa.sendv_sndinfo.snd_context = 0;
+			sendv_spa.sendv_sndinfo.snd_assoc_id = 0;
+
+			sendv_spa.sendv_prinfo.pr_policy = 0;
+			switch (policy) {
+			case 0:
+#ifdef SCTP_PR_SCTP_NONE
```
tuexen, line 833: Same inconsistency as above. Please fix.

Issue #376, opened 2019-09-19
Missing return in error path in sctp_handle_asconf()


markwo:
https://github.com/sctplab/usrsctp/blob/8ca132191668c1dff35872315cf399cd7c3dba9b/usrsctplib/netinet/sctp_asconf.c#L722

tuexen:
Fixed in https://github.com/sctplab/usrsctp/commit/ac364403c13c5cefe2bc5e87b8bf730ba7d491a0. Thanks for finding and reporting!

PR #486, opened 2020-06-19
Adding tests for ASCONF chunks

yuquanw2:
Adding a fuzzing coverage for ASCONF chunks, referring to a previous [issue](https://github.com/sctplab/usrsctp/issues/376).


tuexen:
This PR seems to bring back a feature I took out years ago. While working on [RFC 5061](https://tools.ietf.org/html/rfc5061), there was a time where some implementations had already implemented support for ASCONF chunks, but the authentication was not yet specified or implemented. So for interop testing it was necessary to communicate with implementation not supporting [RFC 4895](https://tools.ietf.org/html/rfc4895). But this is about 13 years ago. Therefore similar code was taken out of the tree and I don't want to being that back.

Why do you need the proposed change? You want to inject ASCONF chunks without getting the AUTH parameters and computation correct. This is already possibly by defining `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`. The result is that any AUTH chunk is accepted. See [sctp_auth.c:1650](https://github.com/sctplab/usrsctp/blob/master/usrsctplib/netinet/sctp_auth.c#L1650). The fuzzer only needs to inject an arbitrary AUTH chunk followed by an ASCONF chunk and it should work. Doesn't this work?

PR #487, opened 2020-06-22
Adding tests for ASCONF chunks

yuquanw2:
Adding fuzzer coverage referring to the [issue](https://github.com/sctplab/usrsctp/issues/376) and a closed [PR](https://github.com/sctplab/usrsctp/pull/486).


fuzzer/fuzzer_connect.c
```diff
@@ -151,6 +151,7 @@ initialize_fuzzer(void) {
 
 	usrsctp_register_address((void *)1);
 	usrsctp_sysctl_set_sctp_pktdrop_enable(1);
+	usrsctp_sysctl_set_sctp_asconf_enable(1);
```
tuexen, line 154: I guess you also want to enable support for SCTP authentication by adding before this line:
```
usrsctp_sysctl_set_sctp_auth_enable(1);
```

yuquanw2: I think that the sctp_auth is set to 1 by default so that I skipped it. Do you think that it will be better to add it to the code?

tuexen: > I think that the sctp_auth is set to 1 by default so that I skipped it. Do you think that it will be better to add it to the code?

Isn't it the same true for sctp_asconf?

yuquanw2: Oops, then I guess I only need to add a CORPUS. Sorry about this.

tuexen: There is no problem in explicitly enabling what you want to enable. That way you are not dependent on the defaults. Just be consistent. Either enable auth and asconf to be sure, or none and depend on the defaults...

As a potential next step you might want to enable NR-SACKs. Using that would make sense for WebRTC. The support is implemented and should be working, it is just not standardised by the IETF...

yuquanw2: Thank you! I will check on that part.

PR #490, opened 2020-06-25
tsctp: add parameter to get the current throughput periodically

msvoelker:
Adds the parameter -d to be able to specify an update time x. If specified, the receiver prints every x seconds the throughput of the last x seconds.


tuexen:
Please fix the builds.

PR #494, opened 2020-06-26
Clean Meson build file

lgrahl:
Fixes #493 

I guess it makes sense to add a GitHub action for Meson, too, so this doesn't happen. Should I write that?

PR #495, opened 2020-06-26
GitHub actions: CMake is preinstalled. CPU count is fixed.

mstyura:
I've observed problems with existing `GitHub Actions` configuration I've proposed.
1. Sometimes `CMake` installation failed with `API limit error`. It is occurred that `GitHub Actions` runners has pre-installed `CMake` which could be used.
2. Sometimes (in my fork) I've observed `Ninja` installation failure. As long as I've added ninja to get automatic parallelization I believe that can be traded off by using default generator with `CMake` switch `--parallel` with hard coded number of `CPU`s, which is `2` for `GH Actions` runners. 

These changes should improve stability of the build. An maybe duration of builds will be reduced, because now additional SW is installed on runners.

### DIFF
--- a/.github/workflows/build-with-cmake.yml
+++ b/.github/workflows/build-with-cmake.yml
@@ -25,10 +25,6 @@ jobs:
       with:
         path: 'usrsctp_source'
 
-    - uses: seanmiddleditch/gha-setup-ninja@v1
-      with:
-        version: '1.10.0'
-
     - name: Prepare dirs
       shell: bash
       run: |
@@ -47,7 +43,6 @@ jobs:
               -Dsctp_inet=${{ matrix.sctp_inet }} \
               -Dsctp_build_programs=ON \
               -Dsctp_build_fuzzer=OFF \
-              -DCMAKE_GENERATOR=Ninja \
               -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/cmake_install \
               ${GITHUB_WORKSPACE}/usrsctp_source
         
@@ -55,6 +50,7 @@ jobs:
       shell: bash
       run: |
         cmake --build cmake_build \
+              --parallel 2 \
               --config ${{ matrix.cmake_build_type }} \
               --target install \
               --clean-first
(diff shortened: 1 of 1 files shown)


tuexen:
There is one thing I observed, which could be improved. If someone creates a PR, the actions are triggered. Now assume that some the the tests fail and the submitter fixes the issue and updates the PR. I think then the actions are not triggered again. Am I right and could this be improved?

lgrahl:
If you observed this in e.g. #494, I think the reason why it failed the second time (where I force-pushed) is that the source branch got merged and then deleted prior to all tests being run. In GitLab, there's a button for "merge as soon as the tests passed" which prevents this but I don't think GitHub has it.

tuexen:
> If you observed this in e.g. #494, I think the reason why it failed the second time (where I force-pushed) is that the source branch got merged and then deleted prior to all tests being run. In GitLab, there's a button for "merge as soon as the tests passed" which prevents this but I don't think GitHub has it.

I think I observed this with https://github.com/sctplab/usrsctp/pull/490. Did the action ran again?
> 
> Edit: I looked at some of the runs and some of them fail due to `[error]API rate limit exceeded for installation ID 9998573.`. It might make sense to reduce the test matrix a little. It's of course best if all build possibilities are tested but also utopic if done for each commit and each PR. From my experience, what works best is to do an extensive test periodically (e.g. every month) or before a new release instead of on every commit. I don't know if GitHub actions offer something like that.

We do excessive testing on the buildbots. So no need to do all this here. However, the buildbots are only triggered for commits to the master branch, because also binaries are run...

lgrahl:
In #490, it kinda looks like the tests were cancelled. Since I don't know the timeline, it could be that this happened due to a merge before they were complete.

> We do excessive testing on the buildbots. So no need to do all this here. However, the buildbots are only triggered for commits to the master branch, because also binaries are run...

In that case, the tests here should probably just include building with all features and nothing else. I think that is a good compromise.

mstyura:
>  I think then the actions are not triggered again. Am I right and could this be improved?

@tuexen I've tried on dummy merge request https://github.com/sctplab/usrsctp/pull/496 and it seems like `GitHub Actions` is triggered on each commit to PR.

>  I looked at some of the runs and some of them fail due to [error]API rate limit exceeded for installation ID 9998573.

@lgrahl I hope this issue was fixed with this PR. But I'll keep my eye on GH actions next day. If not fixed I have another idea to try.

PR #496, opened 2020-06-26
[do not merge]: Check GitHub Action triggers.

mstyura:
This PR investigates an issue that GH is not triggered when new commits are added.


mstyura:
Seems `GitHub Actions` triggered on each commit.

PR #497, opened 2020-06-26
GitHub Actions: Make CMake builds more vebose

mstyura:
This change will help understand why sometimes build is failed.


tuexen:
Thank you very much!

PR #501, opened 2020-07-08
Add #error if HAVE_SCONN_LEN is not defined.

taylor-b:
Recently we realized that WebRTC was using HAVE_SCONN_LEN to detect whether it needed to set the length field, as all the example programs do, but it wasn't actually being defined by our build configuration: https://bugs.chromium.org/p/webrtc/issues/detail?id=10322

This PR adds an error if "usrsctp.h" is included without HAVE_SCONN_LEN being defined in order to catch the problem easier.

I also needed to update the build files since this creates a chicken and egg problem where the error prevents the dynamic detection from working.

Is it expected that all programs including "usrsctp.h" define HAVE_SCONN_LEN, or is that just a convention? Would it make more sense to have usrsctp.h define HAVE_SCONN_LEN itself?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,10 +126,13 @@ endif ()
 
 set(CMAKE_REQUIRED_INCLUDES "${CMAKE_CURRENT_SOURCE_DIR}/usrsctplib")
 
+# Needed to avoid #error when including usrsctp.h.
+set(CMAKE_REQUIRED_DEFINITIONS -DHAVE_SCONN_LEN)
 check_include_file(usrsctp.h have_usrsctp_h)
 if (NOT have_usrsctp_h)
 	message(FATAL_ERROR "usrsctp.h not found")
 endif ()
+unset(CMAKE_REQUIRED_DEFINITIONS)
 
 check_struct_has_member("struct sockaddr" "sa_len" "sys/types.h;sys/socket.h" have_sa_len)
 if (have_sa_len)
@@ -149,12 +152,14 @@ if (have_sin6_len)
 	add_definitions(-DHAVE_SIN6_LEN)
 endif ()
 
+# Needed to avoid #error when including usrsctp.h.
+set(CMAKE_REQUIRED_DEFINITIONS -DHAVE_SCONN_LEN)
 check_struct_has_member("struct sockaddr_conn" "sconn_len" "usrsctp.h" have_sconn_len)
 if (have_sconn_len)
 	message(STATUS "HAVE_SCONN_LEN")
 	add_definitions(-DHAVE_SCONN_LEN)
 endif ()
-
+unset(CMAKE_REQUIRED_DEFINITIONS)
 
 #################################################
 # COMPILER SETTINGS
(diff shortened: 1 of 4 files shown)


lgrahl:
Seems useful if there are no side effects. :+1:

tuexen:
BSD derived platforms have a length field in the `struct sockaddr`, `struct sockaddr_in`, `struct sockaddr_in6`, ... structures. Therefore I defined the `struct sockaddr_sconn` the same way. Configure scripts need to test for this and defined `HAVE_SA_LEN`, `HAVE_SIN_LEN`, `HAVE_SIN6_LEN`. The names have been in sync between the configure scripts and the code using them. You can write your program the way you want, use your own names, you just need to enforce:
```
#if defined(__APPLE__) || defined(__Bitrig__) || defined(__DragonFly__) || \
    defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
struct sockaddr_conn {
        uint8_t sconn_len;
        uint8_t sconn_family;
        uint16_t sconn_port;
        void *sconn_addr;
};
#else
struct sockaddr_conn {
        uint16_t sconn_family;
        uint16_t sconn_port;
        void *sconn_addr;
};
#endif
```
I see the following issues with this pull request:

1. It enforces a name and way to deal with platform specific structures. Normally this is completely up to the application writer and `usrsctp.h` does allow this.
2. It does not enforce that the constant is undefined on the platforms not using `sconn_len`.
3. It does not cover `struct sockaddr`, `struct sockaddr_in`, `struct sockaddr_in6`, and `struct sockaddr_storage`.

lgrahl:
I'm wondering... I've checked the code and there's only one place where the field actually read and compared from usrsctp. I'll make a couple of assumptions/questions. Let me know if I'm wrong.

If one where to incorrectly not define `HAVE_SCONN_LEN`, the consequence would be that the header defines `sconn_len` but the usrsctp code does not handle it. As far as I can tell that is fine unless the application
    * does rely on `sconn_len` having a correctly assigned value, or
    * casts this to some other `struct sockaddr_*` (not sure why one would do that).

So, I wonder if we need to do anything at all here. Furthermore, if my assumption is correct, Chrome and Firefox who both incorrectly do not define `HAVE_SCONN_LEN`, should be fine because they do neither of the above mentioned pitfalls.

tuexen:
My plan was to be as similar as possible with other networking code. I don't see any `HAVE_*` constant on any system include file. That is why I'm using explicit platform specific defines. You you need to do what you would do when writing platform generic networking code. See for example the code for binding the UDP socket in `ekr_loop.c`:
```
        memset(&sin_c, 0, sizeof(struct sockaddr_in));
        sin_c.sin_family = AF_INET;
#ifdef HAVE_SIN_LEN
        sin_c.sin_len = sizeof(struct sockaddr_in);
#endif
        sin_c.sin_port = htons(client_port);
        sin_c.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
```
It is just platforms defined some structures in a platform depended way. Application writers can deal with this however they want. I would prefer to keep that freedom. For dealing with the problem within the library, I did choose a method: the one you normally use with autotools, which was the first build system used. The other were added by contributions and they followed that way. Not sure if they normally have other ways (other naming conventions, for example) to deal with platform specific structures.

taylor-b:
Understood. It's unfortunate we can't do the equivalent of what autotools with the GN build system, but I think we can manage.

PR #511, opened 2020-07-31
Provide ulp_info pointer on send_callback()

jmillan:


usrsctplib/netinet/sctp_usrreq.c
```diff
@@ -9015,7 +9015,7 @@ register_recv_cb(struct socket *so,
 }
 
 int
-register_send_cb(struct socket *so, uint32_t sb_threshold, int (*send_cb)(struct socket *sock, uint32_t sb_free))
+register_send_cb(struct socket *so, uint32_t sb_threshold, int (*send_cb)(struct socket *sock, uint32_t sb_free, void *ulp_info))
```
tuexen, line 9018: No change needed here. Please undo.

usrsctplib/netinet/sctp_input.c
```diff
@@ -602,7 +602,7 @@ sctp_process_init_ack(struct mbuf *m, int iphlen, int offset,
 			     (inp->send_sb_threshold == 0))) {
 				atomic_add_int(&stcb->asoc.refcnt, 1);
 				SCTP_TCB_UNLOCK(stcb);
-				inp->send_callback(stcb->sctp_socket, sb_free_now);
+				stcb->sctp_ep->send_callback(stcb->sctp_socket, sb_free_now, stcb->sctp_ep->ulp_info);
```
tuexen, line 605: Please use
```
inp->send_callback(stcb->sctp_socket, sb_free_now, inp->ulp_info);
```

usrsctplib/netinet/sctp_indata.c
```diff
@@ -4185,7 +4185,7 @@ sctp_express_handle_sack(struct sctp_tcb *stcb, uint32_t cumack,
 			     (inp->send_sb_threshold == 0))) {
 				atomic_add_int(&stcb->asoc.refcnt, 1);
 				SCTP_TCB_UNLOCK(stcb);
-				inp->send_callback(stcb->sctp_socket, sb_free_now);
+				stcb->sctp_ep->send_callback(stcb->sctp_socket, sb_free_now, stcb->sctp_ep->ulp_info);
```
tuexen, line 4188: Please use
```
inp->send_callback(stcb->sctp_socket, sb_free_now, inp->ulp_info);
```

usrsctplib/netinet/sctp_indata.c
```diff
@@ -4913,7 +4913,7 @@ sctp_handle_sack(struct mbuf *m, int offset_seg, int offset_dup,
 			    (inp->send_sb_threshold == 0))) {
 				atomic_add_int(&stcb->asoc.refcnt, 1);
 				SCTP_TCB_UNLOCK(stcb);
-				inp->send_callback(stcb->sctp_socket, sb_free_now);
+				stcb->sctp_ep->send_callback(stcb->sctp_socket, sb_free_now, stcb->sctp_ep->ulp_info);
```
tuexen, line 4916: Please use
```
inp->send_callback(stcb->sctp_socket, sb_free_now, inp->ulp_info);
```

usrsctplib/netinet/sctp_usrreq.c
jmillan: I really think it's needed, are you sure?

jmillan:
> Please also update Manual.tex and Manual.md.

Done

jmillan:
@tuxen,

In line with this PR, it would be extremely useful having a public method to retrieve the `sb_free` at any time. This way we would have the equivalent to the W3C `bufferedAmount` and `bufferedAmountLow`.

In the case of usrsctp `bufferedAmountLow` event would be the `send_cb`, which fires upon certain free send buffer threshold. `bufferedAmount` would be this new api method that retrieves the free send buffer size.

What do you think about it? Can we have such a public method to retrieve the free send buffer size at any moment?

tuexen:
> @tuxen,
> 
> In line with this PR, it would be extremely useful having a public method to retrieve the `sb_free` at any time. This way we would have the equivalent to the W3C `bufferedAmount` and `bufferedAmountLow`.
> 
> In the case of usrsctp `bufferedAmountLow` event would be the `send_cb`, which fires upon certain free send buffer threshold. `bufferedAmount` would be this new api method that retrieves the free send buffer size.
> 
> What do you think about it? Can we have such a public method to retrieve the free send buffer size at any moment?

For webrtc you would need such a value per data channel, which mean stream in SCTP terms, not at the socket level. Do you really want this at the socket layer?

jmillan:
> For webrtc you would need such a value per data channel, which mean stream in SCTP terms, not at the socket level. Do you really want this at the socket layer?

This information is definitely valuable per data channel rather than per socket, yes.

- Shouldn't `send_cb` include the `stream id` it refers to? `receive_cb` already does?
- Back to my initial question, having `send_cb` include the `stream id` plus having a new public method to retrieve the free send buffer size per stream at any moment would be extremely useful and analogous to the W3C spec.

jmillan:
@tuexen,

Within this PR we can already build a mechanism for avoiding packet drops due to the send buffer being full. Honestly I think having the used/free send buffer per socket is perfectly fine and it's not needed per stream (I guess that would be not even possible anyway).

This is how we achieve this:

- We configure a low `sb_threshold` to get the `send_cb` called frequently. It serves us to have a close to real value of the buffered amount.
- Each time `send_cb` is called, it provides, *after this PR*, the free send buffer size, out of which we get the buffered amount (buffer size - free buffer size).
  -  Such value is the real buffered amount for the mediasoup association for that exact moment.
- Every time `usrsctp_sendv` succeeds, we increase the mediasoup association buffered amount.
  -  This serves us to know the quasi-real buffered amount in usrsctp until `send_cb` is executed again.
- Given a `bufferedAmountLowThreshold` that the user can give as `per stream`, we fire`bufferedamountlow` event accordingly.
  -  The event is fired on _every_ `stream id` (`consumers` in mediasoup jargon) according to their thresholds. API user must know that buffered amount value corresponds to the whole socket and not to a specific consumer. Is the user responsibility to use such information accordingly. 

Summarizing, this provides us with:

* `bufferedAmount` property.
* `bufferedAmountThresHold` value (per consumer) upon which we emit the `bufferedamountlow` event.

Just in case you are interested in how we ([mediasoup](https://mediasoup.org/)) do it, you can check it [here](https://github.com/versatica/mediasoup/pull/442/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.hpp&file-filters%5B%5D=.ts).

Thanks

shiretu:
>send buffer per socket is perfectly fine and it's not needed per stream
I would say no. I'm currently doing that in upper layers, and is not exactly fine. We have a project where the wrtc conn is used with multiple channels, some doing real time streaming of audio/video, some doing API calls. I had to do balancing manually, to avoid head of line blocking. Remember RTMP? That protocol, by design, ensure almost same arrival time of both small and large payloads. I'm doing pretty much the same thing, logically speaking. Overall, balancing the channels is instrumental for our work flow, and I dare to say that I'm not the only one with that problem.

The way I'm doing this currently is to never send more than N bytes per channel, where that N == SCTP MTU, in a round robin fashion for all opened channels. Quite a hassle.

jmillan:
My point is that having a send buffer per socket, and the information of the buffered size at any time in your application, you can decide in your application how you want to distribute such free buffer depending on your needs.

I can't think how can you delegate such decision to usrsctp unless it provides with one independent send buffer to each stream id which will be less optimal space wise.

This is just my opinion.

shiretu:
Having that free space per connection would be immensely helpful. No arguments there. My point was that having that space available, would be nice to just divide it by the number of currently opened channels, and provide the result as part of the API, just as it would be done for overall buffer per connection. That will definitely help, as the upper layer will not be forced to keep global contexts and know, at any given time in any given channel, what are the other siblings channels opened. This is what currently tightly couples some of the components in the upper layers. Weakly coupled subsystems are always more flexible.

I_DATA mode would also help in combination with the per-connection space available, and let the SCTP balance the channels. In fact, if I_DATA would be there, I think it will nicely solve this problem, at least the way I understand it: be able to push N lumps of data of various data, and rest assured that they do not do head of line blocking.

jmillan:
> Having that free space per connection would be immensely helpful

This is easily calculated now after this PR (commented [here](https://github.com/sctplab/usrsctp/pull/511#issuecomment-668403544))

But yes, if usrsctp could provide it then it would be more accurate.

> would be nice to just divide it by the number of currently opened channels

In my opinion such behaviour removes flexibility, because I may want to share the buffer differently among my channels in my app. usrsctp can't assume that I'm willing to distribute the available buffer equally among my channels.

shiretu:
>In my opinion such behaviour removes flexibility, because I may want to share the buffer differently among my channels in my app

The suggestion was made to provide that nugget of info: free space per channel, treating them equally. How the app choses to use that info is a different topic. Might very much decide to ignore it. Providing that info helps in decoupling stuff, at least in my case. And I'm pretty sure in other envs too. Right now, as I have explained, I'm forcing each channel hit some sort of a global function (with respect to the channel itself) returning the number of channels. This makes the code sending stuff on that particular data channel not self sufficient.

Again, is not enforcing anything (no MUSTs), just like reporting the per-connection free space does not enforce anything. Is just valuable information, ready to be leveraged by things that needs it

jmillan:
In mediasoup we notify all our app channels about the *total buffered* size (per SCTP socket) each time it changes, but we could very well notify each of them about Total/Number of channels. [here](https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/Transport.cpp#L2830)

In case this serves you.

shiretu:
We have that covered by abstracting it and offer an event per opened channel. Internally, that's quite a lot of management going on behind the scene, literally monitoring the arriving SACKs, and do lots of counting. Performance is not affected, as this is just a bunch of +/- arithmetics. Periodically, the counters are reset, to eliminate any kind of wrong math fallouts.

```
	/**
	 * Triggered when the channel is ready to send more data. This is triggered
	 * as soon as data can be sent over the channel.
	 * 
	 * @param pWebRTCConnection the connection on which the event occurred
	 * @param id the unique id of the channel on which the event occurred
	 * @param pUserOpaqueData the user data which was associated with the channel
	 * @param freeSpaceSize a hint about the maximum amount of bytes which should
	 * be sent over the channel to achieve load balancing on all currently opened
	 * channels. The sender of data should no exceed this value, otherwise other
	 * channels might/will starve. This value is informational.
	 */
	virtual void SignalChannelDry(WebRTCConnection *pWebRTCConnection, channelid_t id,
			void *pUserOpaqueData, size_t freeSpaceSize) = 0;
```

shiretu:
This is the balancing we have now, when `freeSpaceSize` is respected:

https://drive.google.com/file/d/18cIeQfVtC7f4laD6eq2uSfZrvrg3xkt3/view?usp=sharing

PR #514, opened 2020-08-11
add tests for nr sacks

yuquanw2:


fuzzer/fuzzer_fragment.c
```diff
@@ -9,7 +9,7 @@
 #define FUZZ_B_RESERVED2      		(1 << 1)
 #define FUZZ_B_RESERVED3            (1 << 2)
 #define FUZZ_B_RESERVED4      		(1 << 3)
-#define FUZZ_B_RESERVED5            (1 << 4)
+#define NR_SACK_FALG                (1 << 4)
```
tuexen, line 12: Shouldn't this be `NR_SACK_FLAG`?

fuzzer/fuzzer_fragment.c
```diff
@@ -9,7 +9,7 @@
...
 #define FUZZ_B_RESERVED6         	(1 << 5)
 #define I_DATA_FALG                 (1 << 6)
```
tuexen, line 14: and this `I_DATA_FLAG`?

fuzzer/fuzzer_fragment.c
yuquanw2: In corpus, i set first  two bytes to be chunk type. SCTP_NR_SELECTIVE_ACK is 0x10 and corresponding to 1<<4. Are you suggesting me using a different name?

tuexen: > In corpus, i set first two bytes to be chunk type. SCTP_NR_SELECTIVE_ACK is 0x10 and corresponding to 1<<4. Are you suggesting me using a different name?

I'm only asking whether `FALG` should be `FLAG`? Is this a typo?

yuquanw2: I am so sorrrry. Yes, this is a typo. Thank you for pointing out. LOL

PR #527, opened 2020-09-21
Fix gcc/clang type mismatch warnings on Windows

taylor-b:
On Windows, `ntohl` returns an `unsigned long` instead of `uint32_t`, so it expects a "%lu" print format specifier instead of "%u". Fixed by casting to `uint32_t`.

### DIFF
--- a/programs/client_upcall.c
+++ b/programs/client_upcall.c
@@ -72,7 +72,7 @@ int inputAvailable(void)
 	tv.tv_usec = 0;
 	FD_ZERO(&fds);
 #if defined(_WIN32) && !defined(__MINGW32__)
-	FD_SET(_fileno(stdin), &fds);
+	FD_SET((SOCKET)_fileno(stdin), &fds);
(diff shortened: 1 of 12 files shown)


taylor-b:
To make this extra foolproof I could use `PRIu32`; let me know if you'd like that.

tuexen:
Don't we have that issue in more than one place? Why is the compiler only complaining once?

taylor-b:
This is the only place I've found, but I'm only building the fuzzers, not building anything in programs/.

tuexen:
That explains it. We have a couple of places like this in `programs`. Can you also fix the issues in `programs` by casting to `uint32_t`?

taylor-b:
I found a couple other sources of warnings:

* FD_SET uses the SOCKET type rather than int, so passing int into FD_SET hits -Wsign-compare.
* GetLastError returns a DWORD, which is an unsigned long, again hitting -Wformat

I can put these in separate commits if you like.

tuexen:
Thanks for reporting and fixing!

PR #528, opened 2020-09-22
Fix -Wunreachable-code in programs/

taylor-b:
Many of the programs end in a "while (1) { sleep(1); }" or similar
loop, followed by code to close the socket that will never be executed.

### DIFF
--- a/programs/chargen_server_upcall.c
+++ b/programs/chargen_server_upcall.c
@@ -219,13 +219,5 @@ main(int argc, char *argv[])
 		sleep(1);
 #endif
 	}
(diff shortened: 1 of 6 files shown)


taylor-b:
Discovered this when building the programs to fix -Wformat.

Were there plans to allow these programs to terminate gracefully? If so, would you like me to comment out the unreachable code and leave a comment instead of deleting it?

tuexen:
The code is/was just there to show what kind of cleanup would be needed. These programs are intended to show how to use the API. But we can remove it, of course.

PR #530, opened 2020-09-27
Check for the presence of getrandom.

taylor-b:
It's not present when targeting Android API levels less than 28.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif ()
 
 include(CheckStructHasMember)
+include(CheckFunctionExists)
 include(CheckIncludeFile)
 include(CheckIncludeFiles)
 include(CheckCCompilerFlag)
@@ -119,6 +120,13 @@ if (have_stdatomic_h)
 	add_definitions(-DHAVE_STDATOMIC_H)
 endif ()
 
+set(CMAKE_REQUIRED_INCLUDES "<sys/random.h>")
+check_function_exists(getrandom have_getrandom)
+if (have_getrandom)
+	message(STATUS "have_getrandom")
+	add_definitions(-DHAVE_GETRANDOM)
+endif ()
+unset(CMAKE_REQUIRED_INCLUDES)
 
 #################################################
 # CHECK STRUCT MEMBERS
--- a/configure.ac
+++ b/configure.ac
@@ -144,6 +144,10 @@ AC_CHECK_HEADERS(linux/if_addr.h, [], [], [#include <sys/socket.h>])
 AC_CHECK_HEADERS(linux/rtnetlink.h, [], [], [#include <sys/socket.h>])
 AC_CHECK_HEADERS(netinet/ip_icmp.h, [], [], [#include <netinet/ip.h>])
 
+AC_CHECK_DECL(getrandom,
+              AC_DEFINE(HAVE_GETRANDOM, 1, [Define this if your stack has getrandom.]),,
+              [#include <sys/random.h>])
+
 AC_CHECK_MEMBER(struct sockaddr.sa_len,
                 AC_DEFINE(HAVE_SA_LEN, 1, [Define this if your stack has sa_len in sockaddr struct.]),,
                 [#ifdef HAVE_SYS_TYPES_H
--- a/meson.build
+++ b/meson.build
@@ -114,6 +114,12 @@ if compiler.has_header('stdatomic.h')
     add_project_arguments('-DHAVE_STDATOMIC_H', language: 'c')
 endif
 
+# Feature: getrandom
+have_getrandom = compiler.has_function('getrandom', prefix: '#include <sys/random.h>')
+if have_getrandom
+    add_project_arguments('-DHAVE_GETRANDOM', language: 'c')
+endif
+
 # Feature: sockaddr.sa_len
 prefix = '''
 #include <sys/types.h>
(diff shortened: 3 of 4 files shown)


tuexen:
> It's not present for all versions of Android, and apparently not the

That is a problem. @FelixWilhelm stated that every `getrandom()` works on all modern Linux based system. I hoped that Android would be covered. What function can be used on Android instead? I'm also missing ``__Fuchsia__`` and ``__native_client__``? My plan was to add an `#error` on platforms not supporting `arc4random()` or `getrandom()` or `rands()`... I thought this would only affect ``__Fuchsia__`` and ``__native_client__``, which I don't know if these are used at all. But it seems Android versions before 9 are also affected...

> GCE environment used by OSS-fuzz either; see:
> https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25984

I thought I covered https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25984 by https://github.com/sctplab/usrsctp/commit/f4128850b1cffccf45ed6f4ec507d63115ffdbc0.

> 
> Also fixes a mismatch in #ifdefs; "<sys/random.h>" is not included if
> FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION is defined, but
> getrandom was being used either way.

I don't understand this part. If `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined, I think
```
void
read_random(void *buf, size_t size)
{
	memset(buf, 'A', size);
	return;
}
```
is used, which does not use `getrandom()`.

tuexen:
@taylor-b Do you know when the support for Android 8 will end? Restricting support for Android 9 and higher would allow to use `getrandom()`...

felixwilhelm:
Sorry about this. I didn't know Android 8 is missing getrandom().

I asked around and the best replacement for Fuchsia seems to be https://fuchsia.dev/fuchsia-src/reference/syscalls/cprng_draw

Native Client its own randomness API: https://source.chromium.org/chromiumos/chromiumos/codesearch/+/master:src/third_party/libapps/ssh_client/src/file_system.cc;drc=d8bdc1bf1badb9299b84187e23d30d9b4c7de313;l=100

I don't have a dev environment to test these, but I can take a stab at setting something up later this week if no one else volunteers.

tuexen:
> Sorry about this. I didn't know Android 8 is missing getrandom().

No problem. Any suggestion what to do on that platform?

> 
> I asked around and the best replacement for Fuchsia seems to be https://fuchsia.dev/fuchsia-src/reference/syscalls/cprng_draw

Added support for this in https://github.com/sctplab/usrsctp/commit/7f587aee82b46bd1cbb962725ca99e16e2598e99.

> 
> Native Client its own randomness API: https://source.chromium.org/chromiumos/chromiumos/codesearch/+/master:src/third_party/libapps/ssh_client/src/file_system.cc;drc=d8bdc1bf1badb9299b84187e23d30d9b4c7de313;l=100
> 

I'm not familiar with C++, but I don't see how to implement `read_random()` in C using the information provided. Any help welcome!

> I don't have a dev environment to test these, but I can take a stab at setting something up later this week if no one else volunteers.

If you can help with Android 8 and Native Client, I will use an `#error` on the other platforms. Then we should have a better random number generator support on the platforms I know and others will hopefully open an issue...

taylor-b:
> I don't understand this part. If `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined, I think...

You're right, I didn't notice the start of the `#if`/`#elif` sequence.

> Do you know when the support for Android 8 will end?

I don't know; we still have builders targeting Lollipop though, so I assume it won't be soon.

> What function can be used on Android instead? 

It looks like the fallback being used by chromium and boringssl is reading from "/dev/urandom". Is that doable? Presumably you'd want to open the file descriptor in `usrsctp_init` since it's comparatively expensive.

tuexen:
> > I don't understand this part. If `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined, I think...
> 
> You're right, I didn't notice the start of the `#if`/`#elif` sequence.

OK.

> 
> > Do you know when the support for Android 8 will end?
> 
> I don't know; we still have builders targeting Lollipop though, so I assume it won't be soon.

OK. Just wanted to check for an easy way out. If Google would not provide security fixes anymore for
Android 8, we could just drop support for it. But this doesn't work.

> 
> > What function can be used on Android instead?
> 
> It looks like the fallback being used by chromium and boringssl is reading from "/dev/urandom". Is that doable? Presumably you'd want to open the file descriptor in `usrsctp_init` since it's comparatively expensive.

I can open it in `init_random()`. We would need to add a `finish_random()` to close it. Is there an easy way using `#if defined` that we compile for Android and the API level is less than 28?

taylor-b:
> Is there an easy compile time check using `#if defined` that we compile for Android and the API level is less than 28?

Yes, you should be able to use `#if defined(__ANDROID__ ) && __ANDROID_API__ < 28`; both should be defined by the toolchain if applicable.

tuexen:
> > Is there an easy compile time check using `#if defined` that we compile for Android and the API level is less than 28?
> 
> Yes, you should be able to use `#if defined(__ANDROID__ ) && __ANDROID_API__ < 28`; both should be defined by the toolchain if applicable.

OK, thanks. I committed code for Android, which should work. Can you test it? I don't have any way of testing a build for Android...

taylor-b:
It's working, just one remaining issue: https://github.com/sctplab/usrsctp/pull/531

Thanks for the quick turnaround!

PR #533, opened 2020-10-01
Call getrandom via syscall directly on Linux.

taylor-b:
Doing this because getrandom wasn't added to glibc until 2.25.

### DIFF
--- a/usrsctplib/user_environment.c
+++ b/usrsctplib/user_environment.c
@@ -206,7 +206,8 @@ finish_random(void)
 }
 #endif
 #elif defined(__linux__)
-#include <sys/random.h>
+#include <unistd.h>
+#include <sys/syscall.h>
 
 void
 init_random(void)
@@ -222,7 +223,8 @@ read_random(void *buf, size_t size)
 
 	position = 0;
 	while (position < size) {
-		n = getrandom((char *)buf + position, size - position, 0);
+    // Using syscall directly because getrandom isn't present in glibc < 2.25.
+		n = syscall(__NR_getrandom, (char *)buf + position, size - position, 0);
 		if (n > 0) {
 			position += n;
 		}


taylor-b:
Ran into one more issue with `getrandom`; it was added to glibc in version 2.25, but the chromium builders [are using an older version](https://ci.chromium.org/p/chromium/builders/try/linux-rel/503972?).

Also, looking at the [oldest supported versions of popular distros](https://gist.github.com/wagenet/35adca1a032cec2999d47b6c40aa45b1), half are using a version older than 2.25.

Introducing a `HAVE_GETRANDOM` define as I did in the earlier PR would also work for me.

usrsctplib/user_environment.c
```diff
@@ -222,7 +223,7 @@ read_random(void *buf, size_t size)
 
 	position = 0;
 	while (position < size) {
-		n = getrandom((char *)buf + position, size - position, 0);
+		n = syscall(__NR_getrandom, (char *)buf + position, size - position, 0);
```
tuexen, line 226: Can you put in a comment describing why you do the syscall directly. Then I'm happy to merge.

taylor-b: Sorry, I saw this comment after I already pushed

tuexen: > Sorry, I saw this comment after I already pushed

No problem. As long as it works. We are only missing support for __native_client__. Any idea what can be used there?

taylor-b: I don't know much about native client, but I found `nacl_secure_random`: https://source.chromium.org/chromium/chromium/src/+/master:base/rand_util_nacl.cc

tuexen: OK. I can use that and we will see if some reports a bug. I have no way to test it...

saurik:
This doesn't help in the way you are hoping, as those (notably supported) distributions (which includes a number that I'm actively forced to support myself) that are using old versions of glibc are also usually using commensurately old versions of the kernel that then also don't support this syscall. (To be "more honest" about this when doing builds, you can cross-compile to an old distribution to see what it actually has in its kernel headers.) Can you maybe add a !defined() check on __NR_getrandom to the block that is used for old versions of Android and Emscripten, so there's a fallback to /dev/urandom? (In the mean-time I've added a hack locally to use that /dev/urandom block on Linux.)

Issue #147, opened 2017-05-22
Shutdown sequence advice (possible issue?)


shiretu:
Hi,

I need to tunnel the SCTP over DTLS. For that, I am associating _pDTLS pointer to the socket like this:

```
	//store our DTLS, and register the socket
	usrsctp_register_address(_pDTLS);

	//initialize the SCTP connecting address
	_connectAddress.sconn_family = AF_CONN;
	_connectAddress.sconn_port = EHTONS(_port);
	_connectAddress.sconn_addr = _pDTLS;

	//try to bind the socket
	if (usrsctp_bind(_pSCTPSocket, (struct sockaddr *) &_connectAddress,
			sizeof (struct sockaddr_conn)) != 0) {
		int err = errno;
		FATAL("Unable to connect the SCTP socket: (%d) %s", err, strerror(err));
		return false;
	}
```
This way, when the low level SCTP `conn_output` static callback is called, I get the first argument to be a void * which is in fact my DTLS context. Which is perfect.

When SCTP/DTLS dies or gets closed explicitly, I do this:

```
	//unregister the ULP
	usrsctp_deregister_address(_pDTLS);

	//close the socket
	if (_pSCTPSocket != NULL) {
		usrsctp_set_ulpinfo(_pSCTPSocket, NULL);
		usrsctp_shutdown(_pSCTPSocket, SHUT_RDWR);
		usrsctp_close(_pSCTPSocket);
		_pSCTPSocket = NULL;
	}
```

Here is the issue now:
When the `conn_output` callback is called, I still get the address as being a corpse pointer to _pDTLS, despite the fact I have called `usrsctp_deregister_address` and `usrsctp_set_ulpinfo` with NULL. That generates a crash sometimes, because I will try to call stuff from _pDTLS in the destructor part, which is bad (calling pure virtual functions for example). I can protect against this situation, but the code is quite ugly.

Ideally, when `usrsctp_deregister_address(_pDTLS);` is called, all the internal structures holding a `void *` equal with the provided argument should be set to NULL. Basically, all pending stuff in internal queues. After usrsctp_deregister_address/usrsctp_set_ulpinfo is called, there should be absolutely no call made to the `conn_output` callback for that particular address, or at least the provided pointer should be NULL, so a simple `if` statement can protect against bad things.

shiretu:
[...]

PR #535, opened 2020-10-11
Provide a destroy function for registered addresses

sancane:
This patch tries to fix the problem of usrsctp calling callbacks even
after the usrsctp_close function is invoked.
Usrsctp seems to keep the asociation around until the teardown procedure
is completed. The problem is that sometimes it is not possible to
complete the shutdown procedure if the lower transport is gone like SCTP
running on top of DTLS. In such case, callbacks triggered from usrsctp
are providing an pointer address to an applications which could have
been deallocated.
To let applications know when they can safely deallocate memory
registered with an association, we are storing a destroy function which
will be used for usrsctp to notify applications when the association is
gone.
Related issues:
https://github.com/sctplab/usrsctp/issues/405
https://github.com/sctplab/usrsctp/issues/147

### DIFF
--- a/usrsctplib/netinet/sctp_pcb.c
+++ b/usrsctplib/netinet/sctp_pcb.c
@@ -334,6 +334,13 @@ sctp_free_ifa(struct sctp_ifa *sctp_ifap)
 		if (sctp_ifap->ifn_p) {
 			sctp_free_ifn(sctp_ifap->ifn_p);
 		}
+		if (sctp_ifap->destroy_address) {
+			/* sctp_ifap->address is struct sockaddr_conn */
+			if (sctp_ifap->address.sconn.sconn_addr) {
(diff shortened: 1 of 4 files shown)


tuexen:
Does it fix the issue for you? If that is the case, I can try to integrate this with some tweaks to minimise the impact on the kernel implementation...

sancane:
It seems to perform well in all tests I've been doing. I'm sending this patch as proposal addressed to fix this problem in the hope that people more familiar with sctp code who are also suffering this issue can inspect it and look for possible side effects.
Honestly, I don't know usrsctp code very well and I don't understand much of what happens in there, so I'd rather someone with more knowledge than me could have a look at it. I would not like to break something unintentionally (usrsctp code is kind of complicated).

PR #549, opened 2020-11-10
Support building with emscripten.

taylor-b:


taylor-b:
emscripten is a toolchain that compiles to WebAssembly. Previously this was working for us because we were using `__Userspace_os_Linux`, which was since removed. I could check for `__unix__` instead of `__EMSCRIPTEN__`, but I wasn't sure whether the presence of `__unix__` guarantees that `s6_addr8`/`s6_addr16`/`s6_addr32` are defined.

PR #557, opened 2020-12-15
Fall back to using urandom on Linux if getrandom isn't available.

taylor-b:
@mohatry, @saurik, this should address your concerns. Tested on Ubuntu 14.04 (kernel 3.13), with both the "__NR_getrandom not defined" and "__NR_getrandom defined but fails at runtime" code paths.

### DIFF
--- a/usrsctplib/user_environment.c
+++ b/usrsctplib/user_environment.c
@@ -30,10 +30,6 @@
 
 /* __Userspace__ */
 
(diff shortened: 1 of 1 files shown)


tuexen:
Can you check the build failures?

taylor-b:
@tuexen Fixed; it was just that sctp_pcb.h was transitively including stdlib.h.

PR #558, opened 2020-12-16
Add optional SCTP_USE_OPENSSL_RAND define to use openssl's RNG.

taylor-b:
openssl's random number generation code is more robust, and in the case
of building with chromium, is able to bypass sandbox restrictions that
prevent opening /dev/urandom.

### DIFF
--- a/usrsctplib/user_environment.c
+++ b/usrsctplib/user_environment.c
@@ -90,6 +90,36 @@ finish_random(void)
 {
 	return;
 }
+/* This define can be used to optionally use OpenSSL's random number utility,
+ * which is capable of bypassing the chromium sandbox which normally would
+ * prevent opening files, including /dev/urandom.
+ */
+#elif defined(SCTP_USE_OPENSSL_RAND)
+#include <openssl/rand.h>
+
+/* Requiring BoringSSL because it guarantees that RAND_bytes will succeed. */
+#ifndef OPENSSL_IS_BORINGSSL
+#error Only BoringSSL is supported with SCTP_USE_OPENSSL_RAND.
+#endif
+
+void
+init_random(void)
+{
+	return;
+}
+
+void
+read_random(void *buf, size_t size)
+{
+	RAND_bytes((uint8_t *)buf, size);
+	return;
+}
+
+void
+finish_random(void)
+{
+	return;
+}
 #elif defined(__FreeBSD__) || defined(__DragonFly__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__APPLE__) || defined(__Bitrig__)
 #include <stdlib.h>
 


usrsctplib/user_environment.c
```diff
@@ -82,6 +86,34 @@ read_random(void *buf, size_t size)
...
+}
+
+void
+read_random(void *buf, size_t size)
+{
+	if (!RAND_bytes((uint8_t *)buf, size)) {
+#ifdef INVARIANTS
+		panic("RAND_bytes failed.");
+#else
+		abort();
```
taylor-b, line 111: I don't see abort called outside of an INVARIANTS block anywhere else, but failing to retrieve a random number seems severe enough a reason to abort.

This shouldn't normally happen, as openssl seeds the random number generator automatically by default, but it seemed right to remain cautious.

taylor-b (commented):
This is needed for chromium, as opening /dev/urandom from within the sandbox will fail. Note that there is already a similar `SCTP_USE_OPENSSL_SHA1 ` define.

Also, note that this conflicts with my other pull request, so they'll have to be merged one at a time.

usrsctplib/user_environment.c
tuexen: Hmm. I asked initially a couple of consumers of the lib and the result was that the lib should not terminate the program (or write anything to stdout/stderr). The other implementations of `read_random()` retry until they are successful. Can't you do the same here?

taylor-b: All right, got rid of the `abort`.

usrsctplib/user_environment.c
```diff
@@ -85,6 +85,31 @@ read_random(void *buf, size_t size)
 	return;
 }
 
+void
+finish_random(void)
+{
+	return;
+}
+#elif defined(SCTP_USE_OPENSSL_RAND)
```
MirkoBonadei, line 93: Is there a place where this macro needs to be documented? (e.g. maybe https://github.com/sctplab/usrsctp/blob/master/Manual.md)

usrsctplib/user_environment.c
```diff
@@ -85,6 +85,31 @@ read_random(void *buf, size_t size)
...
+	return;
+}
+
+void
+read_random(void *buf, size_t size)
+{
+	while (!RAND_bytes((uint8_t *)buf, size)) {
+#ifdef INVARIANTS
+		panic("RAND_bytes failed.");
+#else
```
tuexen, line 108: I guess this needs to read `#endif` instead of `#else`.

MirkoBonadei, line 108: This should be `#endif`.

usrsctplib/user_environment.c
```diff
@@ -85,6 +85,31 @@ read_random(void *buf, size_t size)
...
+void
+init_random(void)
+{
+	return;
+}
+
+void
+read_random(void *buf, size_t size)
+{
+	while (!RAND_bytes((uint8_t *)buf, size)) {
```
MirkoBonadei, line 105: Do we expect RAND_bytes to succeed on subsequent invocations if the first one didn't return 1?

usrsctplib/user_environment.c
```diff
@@ -85,6 +85,31 @@ read_random(void *buf, size_t size)
...
+#ifdef INVARIANTS
+		panic("RAND_bytes failed.");
+#else
+	}
+	return;
```
MirkoBonadei, line 110: If this function doesn't abort() in case it detects a random number generator issue (when INVARIANTS is not defined), I think it should provide the caller a way to detect the problem (e.g. via return value?).

As of now it silently fails if RAND_bytes returns -1 (https://www.openssl.org/docs/man1.1.1/man3/RAND_bytes.html) which is a failure mode.

MirkoBonadei (changes requested):

usrsctplib/user_environment.c
tuexen: I think it is used internally by Google. If someone provides a patch for documentation / build systems, I'm willing to merge, but I guess this will not be tested by the buildbots.

usrsctplib/user_environment.c
tuexen: I have no idea. In the FreeBSD sources `arc4random` doesn't fail, therefore there is no code to deal with an error case. I'm not sure if there is always a simple way to signal this to the user, since it might occur during packet handling.

usrsctplib/user_environment.c
tuexen: As I said: the model is that the library does not terminate the program...

usrsctplib/user_environment.c
MirkoBonadei: Taylor, can you add a few lines in the documentation under build systems? (I guess here? https://github.com/sctplab/usrsctp/blob/master/Manual.md#building-the-library-and-the-applications).

usrsctplib/user_environment.c
MirkoBonadei: Yes, that might be tricky (I am not an expert of the library) and since this is not an API function for usrsctp this might even require a multi-layer error status propagation.

Taylor: I think this version of `read_random` cannot assume that `RAND_bytes` never fails (like it is documented for `arc4random` so we need to find another solution (see my comment below, -1 is an error condition that will silently return in this case).

usrsctplib/user_environment.c
taylor-b: I feel like documentation for these macros would be out of place there, since it doesn't talk about any build arguments, let alone those with niche usage.

I could add it as an argument in configure.ac/CMakeLists.txt/etc. though; then it would have at least as much documentation as other arguments. What do you think?

usrsctplib/user_environment.c
taylor-b: An infinite loop doesn't seem much preferable to terminating, but if that's what you're doing everywhere else, makes sense to be consistent.

usrsctplib/user_environment.c
taylor-b: Keep in mind this is probably only going to be used by chromium, where we use boringssl and `RAND_bytes` never fails.

MirkoBonadei: Yes, on BoringSSL `RAND_bytes` never fails so if we can enforce BoringSSL here it would be nice (to ensure that users of the standalone WebRTC library don't define `SCTP_USE_OPENSSL_RAND` and also `rtc_ssl_root` in GN.

What do you think about using `OPENSSL_IS_BORINGSSL` (see https://source.chromium.org/chromium/chromium/src/+/master:third_party/boringssl/src/include/openssl/base.h;l=182;drc=ca058c06474c4492f0fbd517679793234dde6165)?

Something like:

```
#include <openssl/base.h>

#ifndef OPENSSL_IS_BORINGSSL
#error Only BoringSSL is supported with SCTP_USE_OPENSSL_RAND.
#endif
```

usrsctplib/user_environment.c
MirkoBonadei: Yes, that sounds good to me.

Or if you think that is too much for a niche macro like this one, let's add a few lines here to explain it (I just think we should not just use it but at least write something about it).

usrsctplib/user_environment.c
MirkoBonadei: I think the infinite loop should be avoided. And it is not used everywhere (e.g. see the Fuchsia version https://github.com/sctplab/usrsctp/blob/c65538fbe14e0141db5053bc24d7cf7d8e2ed8fc/usrsctplib/user_environment.c#L312 or the FreeBSD one https://github.com/sctplab/usrsctp/blob/c65538fbe14e0141db5053bc24d7cf7d8e2ed8fc/usrsctplib/user_environment.c#L103).

If you `#ifndef` on `OPENSSL_IS_BORINGSSL ` (see the other comment), maybe you can remove the loop and call directly like it happens for other flavours?

usrsctplib/user_environment.c
tuexen: Fine with me. One add  comment why it is enforced.

usrsctplib/user_environment.c
tuexen: Makes sense to me...

taylor-b: Done.

usrsctplib/user_environment.c
taylor-b: Done.

usrsctplib/user_environment.c
taylor-b: Done.

taylor-b:
I went with @MirkoBonadei's suggestions (thanks for looking at it!).

MirkoBonadei (approved):
LGTM.

Thanks!

usrsctplib/user_environment.c
davidben: In terms of infinite loops, note the other `read_random` functions all have infinite loop issues. E.g. if opening /dev/urandom fails, `read_random` will just loop forever trying to read from -1. In general, with the exception of EINTR, I wouldn't expect retrying the operation to make progress. That means the library can either abort or return an error up to the caller.

If this project doesn't like to abort, the other option would be making `read_random` return an error. Then everything up the call stack would need to handle that error.

tuexen: > In terms of infinite loops, note the other `read_random` functions all have infinite loop issues. E.g. if opening /dev/urandom fails, `read_random` will just loop forever trying to read from -1. In general, with the exception of EINTR, I wouldn't expect retrying the operation to make progress. That means the library can either abort or return an error up to the caller.
> 
> If this project doesn't like to abort, the other option would be making `read_random` return an error. Then everything up the call stack would need to handle that error.

I'm happy to accept patches, but all changes should be `#ifdef`ed in a way that only the userland stack is affected, since in the kernel the function can not fail. You might need to drop packets if the problem happens in the context of handling received packets...

PR #562, opened 2021-01-04
fix of incorrect type conversion.

ihsinme:
used transform (uint64_t) (srtt * srtt) is incorrect. first, the product will be counted in uint32_t, and only then the conversion to uint64_t will be performed.

I couldn't find the importance of preemptive execution (srtt * srtt), so I suggest a simple fix.
however, if the sequence of the product in the expression ((uint64_t) net-> mtu * (uint64_t) (srtt * srtt)) is important, then I can correct the commit to any of the options:
1. ((uint64_t) srtt * srtt * net-> mtu)
2. ((uint64_t) net-> mtu * ((uint64_t) srtt * srtt))

I would also pay attention to the comparison if (srtt> 0), given the unsigned nature of srtt, I would suggest replacing it with if (srtt != 0).


ihsinme:
good day.
tell me if I still need to do something in this PR.

tuexen:
No, it's fine. Just need to contemplate what the code should be...

PR #563, opened 2021-01-06
Add __declspec(noreturn) to terminate_non_graceful() on Windows

taylor-b:
Informs the compiler that the function never returns, which fixes a -Wsometimes-uninitialized warning.

### DIFF
--- a/usrsctplib/user_environment.h
+++ b/usrsctplib/user_environment.h
@@ -92,7 +92,7 @@ extern u_short ip_id;
 #include <stdlib.h>
 
 #if defined(_WIN32)
-static inline void
+static inline void __declspec(noreturn)
 #else
 static inline void __attribute__((__noreturn__))
 #endif


tuexen:
I don't understand the warning. In the `default` case we are either `panic()`ing or performing `goto out`. In both cases we are not using the value of `did-output`? So what is the issue here?

Is the whitespace change intended?

taylor-b:
Apparently when compiling with `INVARIANTS `, one of our toolchains isn't smart enough to detect that `panic` will always terminate the process. I could also just remove the `goto` from the `#else`.

The white space change wasn't intentional, just something my editor was configured to do by default; fixed.

usrsctplib/netinet/sctputil.c
```diff
@@ -2162,6 +2162,7 @@ sctp_timeout_handler(void *t)
 		did_output = false;
 		break;
 	default:
+		did_output = false;
```
MirkoBonadei, line 2165: What about initializing this at 1782 and let different cases to switch it as needed?

There is a similar warning at line 2609 (`to_ticks`).

MirkoBonadei (commented):
It looks like clang-cl is not able to understand that the default branch, when `INVARIANTS` is defined will crash the process while clang is able to see that and doesn't emit the dianostic. I will create a bug to follow up on this from the compiler point of view.

That said, is it possible to always initialize this variable (and the other one I mentioned on the other comment) or there are drawbacks in doing it?

MirkoBonadei:
Ok, after some investigation it turns out that a `__declspec(noreturn)` is needed here https://github.com/sctplab/usrsctp/blob/97667bb08e1a18913c604eafb1092d7da623032b/usrsctplib/user_environment.h#L95.

It should be:

```
#if defined(_WIN32)
static inline void __declspec(noreturn)
#else
static inline void __attribute__((__noreturn__))
#endif
terminate_non_graceful(void) {
	abort();
}
```

This will help clang-cl to understand that `terminate_non_graceful` doesn't return and it is able to avoid to emit the diagnostic.

Taylor, can you make the change directly in this pull request and revert the initialization? Even if initializing with a default value is better and less error prone, so maybe the fix is to do both.

tuexen:
> Ok, after some investigation it turns out that a `__declspec(noreturn)` is needed here
> 
> https://github.com/sctplab/usrsctp/blob/97667bb08e1a18913c604eafb1092d7da623032b/usrsctplib/user_environment.h#L95
> 
> .
> It should be:
> 
> ```
> #if defined(_WIN32)
> static inline void __declspec(noreturn)
> #else
> static inline void __attribute__((__noreturn__))
> #endif
> terminate_non_graceful(void) {
> 	abort();
> }
> ```
> 
> This will help clang-cl to understand that `terminate_non_graceful` doesn't return and it is able to avoid to emit the diagnostic.
> 
> Taylor, can you make the change directly in this pull request and revert the initialization? Even if initializing with a default value is better and less error prone, so maybe the fix is to do both.

I would suggest to only add the declaration.

taylor-b:
Added `__declspec(noreturn)`, thanks for the suggestion.

PR #566, opened 2021-01-26
 Simplify #ifdefs around pthread typedefs.

taylor-b:


taylor-b:
Suggested by @MirkoBonadei, may make build errors more obvious.

Would it be appropriate to fall into an `#error "Unsupported platform"` if the userland types aren't defined? I notice that's only done in one other place.

tuexen:
This breaks compilation on MacOS X platforms.

usrsctplib/netinet/sctp_os_userspace.h
```diff
@@ -285,11 +285,11 @@ typedef char* caddr_t;
 #include <sys/socket.h>
 #if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__linux__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__native_client__) || defined(__Fuchsia__) || defined(__EMSCRIPTEN_PTHREADS__)
```
MirkoBonadei, line 286: With regards to the errors on macOS, it looks like the `#if` line should also do `|| defined(__APPLE__)` since it is fine to `#include <pthread.h>` on Apple platforms.

It looks like on Apple platforms, `<pthread.h>` is transitively included before reaching this part of the file (to be 100% we should look the preprocessed output) so even if the `#include` gets removed, the `typedef`s work.

MirkoBonadei:
> Suggested by @MirkoBonadei, may make build errors more obvious.
> 
> Would it be appropriate to fall into an `#error "Unsupported platform"` if the userland types aren't defined? I notice that's only done in one other place.

Yes, I think this is a good thing to do (it makes things more explicit, avoiding to partially support platforms).

taylor-b:
Added `|| defined(__APPLE__)`. Isn't it possible there are other platforms where the same thing happens, though? If so, may be best to leave this alone.

MirkoBonadei:
> Added `|| defined(__APPLE__)`. Isn't it possible there are other platforms where the same thing happens, though? If so, may be best to leave this alone.

Yes, that is definitely possible. I am still in favor of this change because the header is not self contained at the moment so this is an improvement (for example, as of now, removing an `#include` in another file can break compilation because of missing pthread types).

MirkoBonadei (approved):

tuexen:
Why not remove line 286 and line 288? At least according to the Posix standard, you should include `pthread.h`. If we figure out that this breaks a platform, we can use an exception list, which will be shorter.

MirkoBonadei:
> Why not remove line 286 and line 288? At least according to the Posix standard, you should include `pthread.h`. If we figure out that this breaks a platform, we can use an exception list, which will be shorter.

Yes, this should be a simplification (easier to read and understand) and I think it is a good idea to try it. An exception for `__EMSCRIPTEN_PTHREADS__` is probably needed, something like:

```
#else /* !defined(Userspace_os_Windows) */
#include <sys/socket.h>

#if defined(__EMSCRIPTEN__) && !defined(__EMSCRIPTEN_PTHREADS__)
#error "Unsupported build configuration ..."
#endif

#include <pthread.h>

typedef pthread_mutex_t userland_mutex_t;
typedef pthread_cond_t userland_cond_t;
typedef pthread_t userland_thread_t;
#endif
```

tuexen:
> > Why not remove line 286 and line 288? At least according to the Posix standard, you should include `pthread.h`. If we figure out that this breaks a platform, we can use an exception list, which will be shorter.
> 
> Yes, this should be a simplification (easier to read and understand) and I think it is a good idea to try it. An exception for `__EMSCRIPTEN_PTHREADS__` is probably needed, something like:
> 
> ```
> #else /* !defined(Userspace_os_Windows) */
> #include <sys/socket.h>
> 
> #if defined(__EMSCRIPTEN__) && !defined(__EMSCRIPTEN_PTHREADS__)
> #error "Unsupported build configuration ..."
> #endif
> 
> #include <pthread.h>
> 
> typedef pthread_mutex_t userland_mutex_t;
> typedef pthread_cond_t userland_cond_t;
> typedef pthread_t userland_thread_t;
> #endif
> ```

We can do that. However, we don't protect against misconfiguration in other cases... I if you prefer this way, update the PR and I'll merge it.

taylor-b:
Updated according to your suggestions.

Issue #554, opened 2020-11-26
Segfault in "sctp_common_input_processing"


cukiernick:
Hi,
I've encountered a segfault when `SOCK_LOCK` was called with null pointer.
I hope the trace included below will be sufficient to identify the problem.

usrsctp version: `a17109528c75d01f6372d5c30851a639684c6e99`

```
Thread 6 "SCTP/IP6 rcv" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffeeffd700 (LWP 15089)]
__GI___pthread_mutex_lock (mutex=0xe8) at ../nptl/pthread_mutex_lock.c:67
67      ../nptl/pthread_mutex_lock.c: No such file or directory.
(gdb) bt full
#0  __GI___pthread_mutex_lock (mutex=0xe8) at ../nptl/pthread_mutex_lock.c:67
        type = <optimized out>
        __PRETTY_FUNCTION__ = "__pthread_mutex_lock"
        id = <optimized out>
#1  0x00005555555a8953 in sctp_common_input_processing (mm=0x7fffe0000b20, iphlen=0, offset=336, length=336, src=0x7fffeeffc5b0,
    dst=0x7fffeeffc590, sh=0x7fffe011e4f0, ch=0x7fffe011e4fc, compute_crc=1 '\001', ecn_bits=0 '\000', vrf_id=0, port=0)
    at /home/user/cpp/usrsctp_lib-cpp/src/usrsctp_lib/usrsctplib/netinet/sctp_input.c:6028
        high_tsn = 4294958943
        fwd_tsn_seen = 0
        data_processed = 0
        m = 0x7fffe009b6e0
        op_err = 0x1
[...]

PR #577, opened 2021-03-29
Added ACCEPT_LOCK in common_input_processing

cukiernick:
https://github.com/sctplab/usrsctp/issues/554
According to the locking key in user_socketvar.h so_head should be
protected with ACCEPT_LOCK


tuexen:
You are right. I merged an equivalent PR. Sorry for taking so long.

PR #581, opened 2021-04-28
Correct option type for UDP encapsulation port

caleb-allen:
It seems to me that the type for SCTP_REMOTE_UDP_ENCAPS_PORT is checked as the struct, `sctp_udpencaps`, and not an `int` as described in the documentation.

https://github.com/sctplab/usrsctp/blob/279a68fd6f5b8b913bd19ca99342f5f3e285a6fe/usrsctplib/netinet/sctp_usrreq.c#L7346


tuexen:
Thanks, you are right.

Issue #580, opened 2021-04-26
The lightweight ip support.


ycyang1229:
HI, 

I am focusing this lib runs on embedded devices which are rtos based platform not linux based, so they may be quipped with lwIP or other popular tcp/ip stack. I think lwIP should be the most popular one. Do you have the interest in merging the implementation of lwip if I add one for this lib?

thanks.

tuexen:
I can't really say without seeing how this impacts the code...

ycyang1229:
> I can't really say without seeing how this impacts the code...

Indeed, let me make one rough version and come back to you.

PR #583, opened 2021-05-03
add lwip support.

ycyang1229:
This is one rough version of lwip support, and it does not support ipv6. lwIP does not support sctp raw socket, so it only can support the feature of udp socket. The macro of ip header and udp header are added for the different naming between linux-based platform and lwip, but I noticed the similar data structure are also added for windows-based platform. Not sure which way is more suitable, so I added the macro first. thanks.

https://github.com/sctplab/usrsctp/issues/580

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,12 @@ option(sctp_sanitizer_memory "Compile with memory sanitizer" 0)
 
 option(sctp_build_fuzzer "Compile in clang fuzzing mode" 0)
 
+option(sctp_use_lwip "build with lwip" ON)
+
+if(sctp_use_lwip)
+	add_definitions(-DSCTP_USE_LWIP)
+endif()
+
 if (sctp_sanitizer_address AND sctp_sanitizer_memory)
 	message(FATAL_ERROR "Can not compile with both sanitizer options")
 endif ()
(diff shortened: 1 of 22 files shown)


CMakeLists.txt
```diff
@@ -79,6 +79,12 @@ option(sctp_sanitizer_memory "Compile with memory sanitizer" 0)
 
 option(sctp_build_fuzzer "Compile in clang fuzzing mode" 0)
 
+option(sctp_use_lwip "build with lwip" ON)
```
weinrank, line 82: Please don't include LWIP by default.

tuexen:
OK, I see what the changes are about. I think getting support for this in will be a  multi step approach. But I need to contemplate a bit how to do that first.
Is LWIP not supported on Windows?

CMakeLists.txt
ycyang1229: Sure, I will disable it when it is merged into it officially. Alright, it is one very rough version to show the idea.

ycyang1229:
> OK, I see what the changes are about. I think getting support for this in will be a multi step approach. But I need to contemplate a bit how to do that first.
> Is LWIP not supported on Windows?

I am not sure about it supports windows, since I usually run it on embedded devices. Let me search it in order to see it runs on windows or not. come back to you later.

ycyang1229:
It looks like lwip can not run on windows. Probably getting one embedded devices is an easier way.

vikramdattu:
@weinrank @ycyang1229 do you plan to revisit and get this merged?

PR #587, opened 2021-05-12
Unlock TCB whenever returning early from sctp_process_cookie_existing.

taylor-b:
Unless returning after `sctp_abort_an_association`.

PR #588, opened 2021-05-12
Fix one more sctp_process_cookie_existing return path...

taylor-b:
...which needs to unlock `stcb` when returning NULL, when forwarding the return
value of `sctp_process_cookie_new`.


taylor-b (commented):
Please take a look; this is something I overlooked in PR #587 because I was only looking for `return (NULL)`.

EDIT: Just noticed from the PR thread that you're already aware of this.

tuexen:
I committed a similar fix, which was tested by Tolya. Just saw the PR after it.

PR #590, opened 2021-05-17
Add support for older versions of Android

oleavr:
Where `getifaddrs()` isn't available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ include(CheckStructHasMember)
 include(CheckIncludeFile)
 include(CheckIncludeFiles)
 include(CheckCCompilerFlag)
+include(CheckSymbolExists)
 
 #################################################
 # CHECK OPTIONS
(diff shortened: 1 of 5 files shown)


tuexen:
Just curious: Are you actually using a native SCTP stack on the mobile platforms?

oleavr:
> Just curious: Are you actually using a native SCTP stack on the mobile platforms?

Yeah we need to be able to interop with WebRTC data channels in [Frida](https://frida.re) – using libnice to implement ICE, GLib / glib-networking for DTLS, and this awesome library for the SCTP layer 🙂

tuexen:
> > Just curious: Are you actually using a native SCTP stack on the mobile platforms?
> 
> Yeah we need to be able to interop with WebRTC data channels in [Frida](https://frida.re) – using libnice to implement ICE, GLib / glib-networking for DTLS, and this awesome library for the SCTP layer 🙂

But this mans you don't need support for SCTP over IP. You just need support for running it over some lower layer. so you don't need to define INET or INET6... Doesn't this already solve your problem?

oleavr:
> > > Just curious: Are you actually using a native SCTP stack on the mobile platforms?
> > 
> > 
> > Yeah we need to be able to interop with WebRTC data channels in [Frida](https://frida.re) – using libnice to implement ICE, GLib / glib-networking for DTLS, and this awesome library for the SCTP layer 🙂
> 
> But this mans you don't need support for SCTP over IP. You just need support for running it over some lower layer. so you don't need to define INET or INET6... Doesn't this already solve your problem?

Ooh right, good point! So I suppose this PR doesn't make any sense as presumably that entire feature would be broken without `getifaddrs()`, and an actually useful PR would be one that implements a fallback. (And until somebody needs that it's better that the build fails than building a partially broken library.)

Sorry for the noise, and thanks a lot for clarifying! 😅

PR #591, opened 2021-05-21
Add usrsctp_get_timeout()

oleavr:
To improve energy efficiency when integrating with an event loop.

This could also be used internally for the timer thread, to improve on the current behavior where this thread wakes up every 10 ms.


oleavr:
Oops, should obviously be finding the minimum delta, not the maximum. Need more coffee! 😅 Fixing.

oleavr:
Fixed.

ibc:
@tuexen any chance for this PR to be reviewed? It makes lot of sense.

jmillan:
Wow, I didn't see this before. This looks like an epic performance and energy improvement!

jmillan:
@oleavr , would you please provide some detailed description about the improvements this change provides?

@tuexen , friendly ping about this nice enhancement.

jmillan:
> To improve energy efficiency when integrating with an event loop.

Even a better approach, when integrating with an even loop, would be being able to assign an event file descriptor (`eventfd()`) to a usrsctp socket, so `usrsctp` writes into the event fd anytime there is something to read/write and the application can poll the event fd to get notified about such changes made by the library. 

This way there would be an alternative to timers, and the app would be notified by the event loop only when there is something to be done.

Lynnworld:
Would it be possible to merge this pull request, @tuexen ? It appears to be quite beneficial.

PR #596, opened 2021-07-08
Propagate error cause code to notifications

Orphis:
Before this change, sac_error would always be set to 0 in case of errors such as protocol violations.
When we receive an ABORT, the error code is propagated just fine to the notifications, but  when the local client is emitting an ABORT, it would always be 0.

### DIFF
--- a/usrsctplib/netinet/sctputil.c
+++ b/usrsctplib/netinet/sctputil.c
@@ -4587,10 +4587,12 @@ sctp_abort_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 #endif
                        uint32_t vrf_id, uint16_t port)
 {
-	uint32_t vtag;
 #if defined(__APPLE__) && !defined(__Userspace__)
 	struct socket *so;
 #endif
+	struct sctp_gen_error_cause* cause;
+	uint32_t vtag;
+	uint16_t cause_code;
 
 	vtag = 0;
 	if (stcb != NULL) {
@@ -4604,7 +4606,15 @@ sctp_abort_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 	                vrf_id, port);
 	if (stcb != NULL) {
 		/* We have a TCB to abort, send notification too */
-		sctp_abort_notification(stcb, 0, 0, NULL, SCTP_SO_NOT_LOCKED);
+
+		if (op_err != NULL) {
+			/* Read the cause code from the error cause */
+			cause = mtod(op_err, struct sctp_gen_error_cause *);
+			cause_code = ntohs(cause->code);
+		} else {
+			cause_code = 0;
+		}
+		sctp_abort_notification(stcb, 0, cause_code, NULL, SCTP_SO_NOT_LOCKED);
 		/* Ok, now lets free it */
 #if defined(__APPLE__) && !defined(__Userspace__)
 		so = SCTP_INP_SO(inp);
@@ -4696,6 +4706,8 @@ sctp_abort_an_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 #if defined(__APPLE__) && !defined(__Userspace__)
 	struct socket *so;
 #endif
+	struct sctp_gen_error_cause* cause;
+	uint16_t cause_code;
 
 #if defined(__APPLE__) && !defined(__Userspace__)
 	so = SCTP_INP_SO(inp);
@@ -4727,6 +4739,15 @@ sctp_abort_an_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 		}
 		return;
 	}
+
+	if (op_err != NULL) {
+		/* Read the cause code from the error cause */
+		cause = mtod(op_err, struct sctp_gen_error_cause *);
+		cause_code = ntohs(cause->code);
+	} else {
+		cause_code = 0;
+	}
+
 	/* notify the peer */
 	sctp_send_abort_tcb(stcb, op_err, so_locked);
 	SCTP_STAT_INCR_COUNTER32(sctps_aborted);
@@ -4736,7 +4757,7 @@ sctp_abort_an_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 	}
 	/* notify the ulp */
 	if ((inp->sctp_flags & SCTP_PCB_FLAGS_SOCKET_GONE) == 0) {
-		sctp_abort_notification(stcb, 0, 0, NULL, so_locked);
+		sctp_abort_notification(stcb, 0, cause_code, NULL, so_locked);
 	}
 	/* now free the asoc */
 #ifdef SCTP_ASOCLOG_OF_TSNS


Orphis:
Hi @tuexen, any chance we could get this reviewed and merged soon?

tuexen:
> Hi @tuexen, any chance we could get this reviewed and merged soon?

I'm busy working on Internet Drafts until next Monday night (then is the deadline). Can look at it next week? OK?

Orphis:
My deadline is on Thursday next week (next Chrome branching), so the sooner, the better. I guess that Tuesday will be fine and I hope I can address any review comment on the same day.

tuexen:
So we are talking about [SCTP_ASSOC_CHANGE](https://datatracker.ietf.org/doc/html/rfc6458#section-6.1.1). Right now we populate the `sac_error` field only, if we received an ABORT chunk from the peer by putting in the cause code of the first error cause contained in the ABORT chunk (if it exists).

You are proposing to do the same if we send out an ABORT chunk. [SCTP_ASSOC_CHANGE](https://datatracker.ietf.org/doc/html/rfc6458#section-6.1.1) is unclear about this.

@stewrtrs What do you think? Should we also populate this value for outgoing ABORT chunks?

stewrtrs:
Hmm

I suppose it would not hurt.. though its definitely not in the spec ;)

> On Jul 8, 2021, at 2:39 PM, Michael Tüxen ***@***.***> wrote:
> 
> 
> So we are talking about SCTP_ASSOC_CHANGE. Right now we populate the sac_error field only, if we received an ABORT chunk from the peer by putting in the cause code of the first error cause contained in the ABORT chunk (if it exists).
> 
> You are proposing to do the same if we send out an ABORT chunk. SCTP_ASSOC_CHANGE is unclear about this.
> 
> @stewrtrs What do you think? Should we also populate this value for outgoing ABORT chunks?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 

------
Randall Stewart
***@***.***

usrsctplib/netinet/sctputil.c
```diff
@@ -4604,7 +4606,10 @@ sctp_abort_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 	                vrf_id, port);
 	if (stcb != NULL) {
 		/* We have a TCB to abort, send notification too */
-		sctp_abort_notification(stcb, 0, 0, NULL, SCTP_SO_NOT_LOCKED);
+		/* Read the cause code from the abort chunk */
+		cause = mtod(op_err, struct sctp_gen_error_cause *);
```
tuexen, line 4610: This results in dereferencing a `NULL` pointer, if `op_err` is `NULL`. So I suggest to use something like
```
if (op_err != NULL) {
    cause = mtod(op_err, struct sctp_gen_error_cause *);
    cause_code = ntohs(cause->code);
} else {
    cause_code = 0;
}

File: usrsctplib/netinet/sctputil.c [hunk 2]
```diff
@@ -4727,6 +4734,11 @@ sctp_abort_an_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 		}
 		return;
 	}
+
+	/* Read the cause code from the abort chunk */
+	cause = mtod(op_err, struct sctp_gen_error_cause *);
+	cause_code = ntohs(cause->code);
+
 	/* notify the peer */
```
tuexen, line 4742: I guess you need here also:
```
if (op_err != NULL) {
    cause = mtod(op_err, struct sctp_gen_error_cause *);
    cause_code = ntohs(cause->code);
} else {
    cause_code = 0;
}
```

usrsctplib/netinet/sctputil.c
```diff
@@ -4693,6 +4698,8 @@ sctp_abort_an_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
                           struct mbuf *op_err,
                           int so_locked)
 {
+	uint16_t cause_code;
+	struct sctp_gen_error_cause* cause;
```
tuexen, line 4702: Please use:
```
struct sctp_gen_error_cause* cause;
#if defined(__APPLE__) && !defined(__Userspace__)
 	struct socket *so;
#endif
uint16_t cause_code;

File: usrsctplib/netinet/sctputil.c [hunk 4]
```diff
@@ -4588,6 +4588,8 @@ sctp_abort_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
                        uint32_t vrf_id, uint16_t port)
 {
 	uint32_t vtag;
+	uint16_t cause_code;
+	struct sctp_gen_error_cause* cause;
 #if defined(__APPLE__) && !defined(__Userspace__)
 	struct socket *so;
 #endif
```
tuexen, line 4595: Please use:
#if defined(__APPLE__) && !defined(__Userspace__)
     struct socket *so;
#endif
    struct sctp_gen_error_cause* cause;
    uint32_t vtag;
    uint16_t cause_code;

usrsctplib/netinet/sctp_input.c
```diff
@@ -1264,7 +1264,8 @@ sctp_handle_error(struct sctp_chunkhdr *ch,
 				asoc->stale_cookie_count++;
 				if (asoc->stale_cookie_count >
 				    asoc->max_init_times) {
-					sctp_abort_notification(stcb, 0, 0, NULL, SCTP_SO_NOT_LOCKED);
+					sctp_abort_notification(stcb, 0, SCTP_CAUSE_STALE_COOKIE, NULL,
+					              SCTP_SO_NOT_LOCKED);
```
tuexen, line 1268: Hmm. Shouldn't this be handled the same way has too many retransmits?

tuexen:
@Orphis : Are you setting the `net.inet.sctp.diag_info_code` `sysctl`-variable? Without this, you don't provide always an error cause. I guess, security people don't like providing error causes at all. That is why I made these at least controllable via the `sysctl` interface.

tuexen:
> Hmm I suppose it would not hurt.. though its definitely not in the spec ;)

I did some testing on Linux. I used a received INIT-ACK chunk without a cookie parameter as an example. Linux does provide an error code, which should be 7, but actually is 1792. So Linux is missing a byte order conversion...

Here is a `packetdrill` script which currently passes on Linux and FreeBSD and therefore shows the current behaviour:
```
 0.0 socket(..., SOCK_STREAM, IPPROTO_SCTP) = 3
+0.0 fcntl(3, F_GETFL) = 0x2 (flags O_RDWR)
+0.0 fcntl(3, F_SETFL, O_RDWR|O_NONBLOCK) = 0
+0.1 connect(3, ..., ...) = -1 EINPROGRESS (Operation now in progress)
#ifdef FreeBSD
+0.0 setsockopt(3, IPPROTO_SCTP, SCTP_EVENT, {se_assoc_id=0, se_type=SCTP_ASSOC_CHANGE, se_on=1}, 8) = 0
#endif
#ifdef Linux
+0.0 setsockopt(3, IPPROTO_SCTP, SCTP_EVENTS, {sctp_data_io_event=0,
                                               sctp_association_event=1,
                                               sctp_address_event=0,
                                               sctp_send_failure_event=0,
                                               sctp_peer_error_event=0,
                                               sctp_shutdown_event=0,
                                               sctp_partial_delivery_event=0,
                                               sctp_adaptation_layer_event=0,
                                               sctp_authentication_event=0,
                                               sctp_sender_dry_event=0}, 11) = 0
#endif
+0.0 > sctp: INIT[flgs=0, tag=1, a_rwnd=..., os=..., is=..., tsn=1, ...]
+0.1 < sctp: INIT_ACK[flgs=0, tag=2, a_rwnd=1500, os=1, is=1, tsn=1]
+0.0 > sctp: ABORT[flgs=0, MISSING_MANDATORY_PARAMETER[types=[7]]]
+0.0 recvmsg(3, {msg_name(...)=...,
                 msg_iov(1)=[{iov_base={sac_type=SCTP_ASSOC_CHANGE,
                                        sac_flags=0,
                                        sac_length=20,
                                        sac_state=SCTP_CANT_STR_ASSOC,
#ifdef FreeBSD
                                        sac_error=0,
#endif
#ifdef Linux
                                        sac_error=1792,
#endif
#ifdef FreeBSD
                                        sac_outbound_streams=10, 
#endif
#ifdef Linux
                                        sac_outbound_streams=0, 
#endif
                                        sac_inbound_streams=0,
                                        sac_assoc_id=...,
                                        sac_info=[]},
                              iov_len=1000}],
                 msg_control(0)=[],
                 msg_flags=MSG_NOTIFICATION|MSG_EOR}, 0) = 20
#ifdef FreeBSD
+0.0 getsockopt(3, SOL_SOCKET, SO_ERROR, [ETIMEDOUT], [4]) = 0
#endif
#ifdef Linux
+0.0 getsockopt(3, SOL_SOCKET, SO_ERROR, [ECONNREFUSED], [4]) = 0
#endif
```

So I guess we can add this...

usrsctplib/netinet/sctputil.c
Orphis: I moved the `struct sctp_gen_error_cause* cause;` below the `#if` as well.

usrsctplib/netinet/sctp_input.c
Orphis: I removed that part from the PR, if you can think of a better error value to provide, it can be added in a later change I guess.

Orphis:
Updated the commit with suggested changes and removed the COOKIE related one.

```
   sac_error:  If the state was reached due to an error condition (e.g.,
      SCTP_COMM_LOST), any relevant error information is available in
      this field.  This corresponds to the protocol error codes defined
      in [RFC4960].
```

I don't see any conditionality for this field in RFC6458 and it should probably be present most of the time if the notification is emitted. I'm not an expert in that RFC, so please help me understand why it shouldn't be there.

If security researchers believe it shouldn't be there in some cases and using the sysctl is the right way to have it included, I'm not opposed to that. We don't currently set it in Chrome, not did I need it in my limited testing to catch some errors. As I understand the `SYSCTL` related macro, any calls such as `sctp_generate_cause(SCTP_BASE_SYSCTL(sctp_diag_info_code), "...");` will pass a code of `1` (or whatever the value is in `sctp_diag_info_code`) and won't match a cause code defined in the IANA registry. Is this intended behavior?

Note that the error reporting is part of the WebRTC specification as well (see https://w3c.github.io/webrtc-pc/#dom-rtcerror-sctpcausecode ) which is what I'm trying to fix. It's also important for backend services to be able to log the proper error without doing packet inspection and checking the ABORT chunk.

I do agree the value in there should be in host byte order and that Linux is in the wrong there.

tuexen:
> Updated the commit with suggested changes and removed the COOKIE related one.

Great, thanks a lot.

> 
> ```
>    sac_error:  If the state was reached due to an error condition (e.g.,
>       SCTP_COMM_LOST), any relevant error information is available in
>       this field.  This corresponds to the protocol error codes defined
>       in [RFC4960].
> ```
> 
> I don't see any conditionality for this field in RFC6458 and it should probably be present most of the time if the notification is emitted. I'm not an expert in that RFC, so please help me understand why it shouldn't be there.

I think the original intention was to provide a hint in case of an incoming error. This would be in tune with
```
sac_info:  If sac_state is SCTP_COMM_LOST and an ABORT chunk was
      received for this association, sac_info[] contains the complete
      ABORT chunk as defined in Section 3.3.7 of the SCTP specification
      [RFC4960].
```
where it is made explicit. But you are right, that the specification has not the restriction for `sac_error`. But it also does not explicitly mentions the case of an outgoing ABORT chunk. The Linux developers interpreted the text like you and we'll add this code.

> 
> If security researchers believe it shouldn't be there in some cases and using the sysctl is the right way to have it included, I'm not opposed to that. We don't currently set it in Chrome, not did I need it in my limited testing to catch some errors. As I 

Security researchers don't like providing any information to peer. So they would prefer not to send any error causes. We thought it really helps in figuring out what is going on and the security of an implementation should not be based one the non-availabilty of such information, and therefore the FreeBSD stacks sends them.

>understand the `SYSCTL` related macro, any calls such as `sctp_generate_cause(SCTP_BASE_SYSCTL(sctp_diag_info_code), "...");` will pass a code of `1` (or whatever the value is in `sctp_diag_info_code`) and won't match a cause code defined in the IANA registry. Is this intended behavior?

You can set it to any value, preferably to a non-IANA assigned one. Then it is used whenever no other error cause seems appropriate.

> 
> Note that the error reporting is part of the WebRTC specification as well (see https://w3c.github.io/webrtc-pc/#dom-rtcerror-sctpcausecode ) which is what I'm trying to fix. It's also important for backend services to be able to log the proper error without doing packet inspection and checking the ABORT chunk.

I don't understand the WebRTC specification. It needs to define the values and the corresponding semantic. So for example if the association is terminated due to excessive retransmissions, there is no SCTP level cause code for this. So what is provided at the API level to indicate this to the application? Right the WebRTC specification does not make sense to me in regard to this field.

> 
> I do agree the value in there should be in host byte order and that Linux is in the wrong there.

Will report it to them after cleaning up FreeBSD and the `packetdrill` script.

usrsctplib/netinet/sctputil.c
```diff
@@ -4604,7 +4606,15 @@ sctp_abort_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 	                vrf_id, port);
 	if (stcb != NULL) {
 		/* We have a TCB to abort, send notification too */
-		sctp_abort_notification(stcb, 0, 0, NULL, SCTP_SO_NOT_LOCKED);
+
+		if (op_err != NULL) {
+			/* Read the cause code from the abort chunk */
```
tuexen, line 4611: Nit: Can you replace `abort chunk` by `error cause`?

usrsctplib/netinet/sctputil.c
```diff
@@ -4727,6 +4739,15 @@ sctp_abort_an_association(struct sctp_inpcb *inp, struct sctp_tcb *stcb,
 		}
 		return;
 	}
+
+	if (op_err != NULL) {
+		/* Read the cause code from the abort chunk */
```
tuexen, line 4744: Nit: Can you replace `abort chunk` by `error cause`?

tuexen:
@Orphis Could you address the last two comments (nits, but it would be great if you can fix them? Then I'll merge it.

tuexen:
Thanks for bringing this up and proving the patch.

Orphis:
I believe that the WebRTC specification intent is to provide one whenever possible. If there's no sctpCauseCode available, then the field will be empty. The message description can still have a more descriptive text message if one is available.
At the moment in Chrome, it is just a text representation of the cause code name for the usrsctp backend.

tuexen:
> I believe that the WebRTC specification intent is to provide one whenever possible. If there's no sctpCauseCode available, then the field will be empty. The message description can still have a more descriptive text message if one is available.
> At the moment in Chrome, it is just a text representation of the cause code name for the usrsctp backend.

Isn't the type `long` and shouldn't this value describe why an SCTP association was terminated? I think indicating that the peer is not reachable anymore is an important case and should be covered. But I'll leave that to the authors of the WebRTC spec. I'm not familiar with JavaScript APIs and what is expected there or not.

Orphis:
> Isn't the type `long` and shouldn't this value describe why an SCTP association was terminated? I think indicating that the peer is not reachable anymore is an important case and should be covered. But I'll leave that to the authors of the WebRTC spec. I'm not familiar with JavaScript APIs and what is expected there or not.

Yes, `long` is definitely an oversight. So far, we haven't had that feature implemented in any of the WebRTC implementation, so this will give us experience to possibly change it if users don't find it useful in its current form.

tuexen:
@Orphis Please be sure also to include https://github.com/sctplab/usrsctp/commit/fc5cba21a01b60977d277b0e6b5b72146383dcf6. It avoids a use-after-free issue introduced by this commit. `op_err` is freed by `sctp_send_abort()`, so we need to cache the `cause_code` before calling it.

Orphis:
Oh you're right, I did notice that in one of the call sites and fixed it there, but missed the second one. Sorry about that!

tuexen:
No problem, found it right away during testing before committing it to FreeBSD... Fixing `errno` is more extensive changes...

PR #601, opened 2021-07-24
Change defines to type aliases for better compatibility

### LINKED ISSUE #600: Compilation failure on MinWG GCC
I'm getting this error with latest master branch as of right now even though it is supposedly supported from the issues and PRs I found:
```
..\..\subprojects\usrsctp-1ade45cbadfd19298d2c47dc538962d4425ad2dd\usrsctplib/usrsctp.h:69:19: error: expected primary-expression before 'unsigned'
 #define uint32_t  unsigned __int32
                   ^~~~~~~~
..\..\subprojects\usrsctp-1ade45cbadfd19298d2c47dc538962d4425ad2dd\usrsctplib/usrsctp.h:69:19: note: in definition of macro 'uint32_t'
 #define uint32_t  unsigned __int32
                   ^~~~~~~~
```

Meson detects compiler as:
```
gcc 8.1.0 "gcc.exe (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 8.1.0"
```

Here is the job where this error happened in case it is helpful: https://github.com/nazar-pc/mediasoup/runs/3148235269?check_suite_focus=true

Looks like it is somehow related to its usage in linked project as `uint32_t{0}`, replacing it with `(uint32_t){0}` helps, maybe it can be defined as such from the beginning (`#define uint32_t  (unsigned __int32)`)?

nazar-pc:
I do not necessarily know what I'm doing, but I think this fixes #600 and stops colliding with `std::*_t` in C++ projects.

### DIFF
--- a/usrsctplib/usrsctp.h
+++ b/usrsctplib/usrsctp.h
@@ -64,19 +64,19 @@ extern "C" {
 #elif defined(SCTP_STDINT_INCLUDE)
 #include SCTP_STDINT_INCLUDE
 #else
-#define uint8_t   unsigned __int8
-#define uint16_t  unsigned __int16
-#define uint32_t  unsigned __int32
-#define uint64_t  unsigned __int64
-#define int16_t   __int16
-#define int32_t   __int32
+typedef unsigned __int8  uint8_t;
+typedef unsigned __int16 uint16_t;
+typedef unsigned __int32 uint32_t;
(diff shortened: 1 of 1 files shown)


nazar-pc:
Are there any concerns about this PR? It seems like should be trivially mergeable and CI is happy.

PR #640, opened 2021-11-30
GitHub Actions: Add CIFuzz integrations

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which usrsctp already runs.


tuexen:
Let's try it and see how it works...

PR #641, opened 2021-12-15
Mention OpenBSD support

klemensn:
`usrsctp` 0.9.5.0 has been tested/used successfully on
OpenBSD/amd64 7.0 -CURRENT as of today in conjunction with
the official Telegram Desktop client.


tuexen:
Thanks for testing and reporting.

PR #644, opened 2022-01-02
cmake: usrsctplib: link against threads library

klemensn:
The usrsctp library uses threads, but building it with cmake does not
link against the system's thread library.

Use cmake's "Threads" package as usual to deal with all supported
platforms.
(autoconf's configure.ac correctly adds `-pthread` to CFLAGS.)

This is needed because programs which do not use threads themselves but
link against usrsctplib will otherwise fail to link since `pthread_*`
symbols used in libusrsctp.so are not resolved.

programs/CMakeFiles.txt avoids this problem by unconditionally using the
"Threads" package for every program including those that do not use
threads themselves.

### DIFF
--- a/usrsctplib/CMakeLists.txt
+++ b/usrsctplib/CMakeLists.txt
@@ -95,6 +95,8 @@ if (sctp_build_shared_lib)
 	set(BUILD_SHARED_LIBS 1)
 endif ()
 
+find_package(Threads)
+
 #################################################
 # LIBRARY FILES
 #################################################
@@ -187,6 +189,8 @@ target_include_directories(usrsctp PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 if(SCTP_USE_MBEDTLS_SHA1)
 	target_include_directories(usrsctp PRIVATE ${MBEDTLS_INCLUDE_DIRS})
 endif()
+
+target_link_libraries(usrsctp ${CMAKE_THREAD_LIBS_INIT})
 
 if (WIN32)
 	message(STATUS "link library: ws2_32")


tuexen:
@weinrank : What do you think?

weinrank:
The change successfully adds the "-pthread" flag as a compiler option, I'm fine with it.

Do we also need this flag for Apple?

General question, since I've not clue about the pthread flag itself: why do we need it?
Do you face any compiler/linker errors? I'm not arguing against the flag and after googling/reading man pages, I think it's a good idea to add it. I'm just questioning myself why we haven't seen any errors/issues for the past years of omitting the flag.

Best

klemensn:
> The change successfully adds the "-pthread" flag as a compiler option, I'm fine with it.
> 
> Do we also need this flag for Apple?

I don't know, but my udpated diff lets cmake's "Threads" package handle all platform specific bits, so that is no longer our problem.

> General question, since I've not clue about the pthread flag itself: why do we need it? Do you face any compiler/linker errors? I'm not arguing against the flag and after googling/reading man pages, I think it's a good idea to add it. I'm just questioning myself why we haven't seen any errors/issues for the past years of omitting the flag.

The new commit message hopefully explains it better.
tl;dr:  Unless your program linking against `libusrsctp.so` (which does not link against `phtreads` itself) uses `-pthreads` or `-lpthreads`, it will either fail to link at compile-time with something like
```
ld: error: ./usrsctplib/libusrsctp.so: undefined reference to pthread_rwlock_rdlock [--no-allow-shlib-undefined]
[ more undefined references to other pthread_* symbols ]
```
or it'll fail to run eventually:
```
test:/usr/local/lib/libusrsctp.so.0.1: undefined symbol 'pthread_mutexattr_init'                         
ld.so: test_libmgmt: lazy binding failed!                                                                        
Killed
```

tuexen:
OK, so the point is that you want to deal with `-pthread` at the library level, not at the application level as we do today.

If that is correct, then you might want to remove the commands you added to `usrsctplib/CMakeLists.txt` also from `usrsctplib/programs/CMakeLists.txt`. Can you do that? If that still works, I'm happy to merge it.

PR #655, opened 2022-04-06
Simplify init_random() to remove QNX special-case

oleavr:
As it has both device entries and they provide the same functionality.

(I totally missed this when skimming the [documentation](http://www.qnx.com/developers/docs/7.0.0/index.html#com.qnx.doc.neutrino.utilities/topic/r/random.html), and it wasn't until today that I happened to notice the other device entry.)


tuexen:
Thanks for the cleanup.

PR #658, opened 2022-04-12
bugfix: nat-state-update for ipv4

sriramy:
The param type looks wrong for aa_del, correcting the param type used for delete to SCTP_DEL_IP_ADDRESS.


tuexen:
Thanks for catching the bug.

PR #674, opened 2023-05-04
fix build error on windows

AvenSun:
```
sctp_output.c
L:\usrsctplib\netinet\sctp_output.c(14823): error C2220: the following warning is treated as an error
L:\usrsctplib\netinet\sctp_output.c(14823): warning C4703: potentially uninitialized local pointer variable 'inp' used
L:\usrsctplib\netinet\sctp_output.c(14827): warning C4703: potentially uninitialized local pointer variable 'asoc' used
```

### DIFF
--- a/usrsctplib/netinet/sctp_output.c
+++ b/usrsctplib/netinet/sctp_output.c
@@ -13477,10 +13477,10 @@ sctp_lower_sosend(struct socket *so,
 #endif
 	struct timeval now;
 	struct sctp_block_entry be;
-	struct sctp_inpcb *inp;
+	struct sctp_inpcb *inp = NULL;
 	struct sctp_tcb *stcb = NULL;
 	struct sctp_nets *net;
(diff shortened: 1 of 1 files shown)


tuexen:
Can you show me a possibility how the variable can be use uninitialized? Right now my understanding is that this is a false positive of the Windows compiler. Neither gcc nor clang nor other static code analyzers report a problem...

AvenSun:
Hi, @tuexen
actually there's no modification of the code.
Just pull and build as a static library on windows via visual studio 2019. 
image
If we don't initialize these two variables during its declaration, 
visual studio will raise an error and interrupt the compilation process.
It might be a false positive of visual studio.
but I think initialization might be more straightforward compared to change the setting of visual studio.

garbear:
This fixes building on win32/win64/UWP32/UWP64/UWP-arm for me.

PR #684, opened 2023-07-31
Fixed a typo that is breaking builds downstream

fleetwoodstack:
This erroneous 's' is causing build errors downstream when we reference this as one of our dependencies. I've removed it.


garbear:
Seems this was fixed upstream: https://github.com/sctplab/usrsctp/commit/8fef5020752fa2f1d1b4760dd95246fc6c80031f

Compilation now works for me on win32/64/32UWP/64UWP/armUWP. I suggest this PR be closed.

PR #695, opened 2024-01-13
Add MSVC solution/projects, fix many warnings, and small enhancements.

axelriet:
- Added Visual Studio solutions to msvc/ - Not relying on CMake and VS's less-than-perfect CMake integration. Fuzzer building on Windows for the first time.
- Modified programs/programs_helper.h (define __PRETTY_FUNCTION__)
- Modified programs/test_libmgmt.c (reduced iterations so the test does not hang for 10 minutes)
- Modified usrsctplib/netinet/sctp_userspace.c (Added support for thread naming on Windows)
- Modified usrsctplib/user_socket.c (usrsctp_getladdrs()/usrsctp_freeladdrs()/usrsctp_getpaddrs()/usrsctp_freepaddrs() - made sure the output pointer is set to null upon function entry and that the matching free() functions can handle null. Before that change, starting programs/client.c before the server resulted in a heap corruption: the call to usrsctp_getpaddrs() at line 226 returned < 0, leaving addrs untouched. But addrs had been freed at line 224, and was freed again at line 272, corrupting the heap.)
- Added contitional __cdecl attributes to some functions (notably all main() functions) so the library can be built with different calling conventions. Ideally, all public/exported function should have their calling convention specified explicitely in the public header, via some new SCTP_PUBLIC macro that is defined per-compiler and build option, so people can build their app and the library with wathever calling convention they choose. On Windows, we have __fastcall as well as __vectorcall (that everyone should really use) and this change enables building the apps and the lib with those conventions without breaking the CRT or the fuzzer.
- Fixed most /W4 warnings (many type casts) instead of globaly suppressing them.


tuexen:
Could you please split this pull request up in separate ones, each one addressing one issue or adding one feature.

PR #699, opened 2024-01-29
Assert if soref() is called on an already-released socket.

JonathanLennox:
If there are still lingering userspace socket double-free bugs, this should hopefully help catch them.


JonathanLennox:
Unfortunately this doesn't work, because `sonewconn()` creates the socket with a ref count of 0, and the refcount doesn't become 1 until `user_accept`.

PR #701, opened 2024-02-06
Fix detection of sockaddr_conn.sconn_len when srcdir != objdir.

JonathanLennox:


JonathanLennox:
To reproduce -- configure outside your srcdir on a BSD-based platform (including macOS).

E.g.
``
./bootstrap
mkdir obj
cd obj
../configure
make
``

This idiom is common in situations where you want to build for more than one architecture (e.g. x86_64 and arm64 on macOS).

Issue #676, opened 2023-05-11
Aborted with sodealloc(): so_count -58769387 -- suspected use-after-free in sctp_timeout_handler


JonathanLennox:
This is the same environment as https://github.com/sctplab/usrsctp/issues/673, and I suspect the same root cause, but a different crash - perhaps this one will give more information.

I am getting an assert crash in usrsctp with the assertion
```
sodealloc(): so_count -58769387
```
which certainly looks like the socket object has been freed and re-used by another allocation.  In my logs I can see that usrsctp_close was called just before the crash.

The crash is in the sctp_timeout_handler, and the stack is
Thread 1 (Thread 0x7fc75affd640 (LWP 7273)):
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140494201935424) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140494201935424) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140494201935424, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007fc88d788476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007fc88d76e7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007fc760e089c5 in terminate_non_graceful () at ./user_environment.h:100
#6  0x00007fc760e094d5 in sodealloc (so=0x7fc7f4015cf0) at user_socket.c:230
#7  0x00007fc760e09ef4 in sofree (so=0x7fc7f4015cf0) at user_socket.c:302
#8  0x00007fc760eba401 in sctp_timeout_handler (t=0x7fc7c0009538) at netinet/sctputil.c:2218
#9  0x00007fc760e1600c in sctp_handle_tick (elapsed_ticks=10) at netinet/sctp_callout.c:172
#10 0x00007fc760e16258 in user_sctp_timer_iterate (arg=0x0) at netinet/sctp_callout.c:214
#11 0x00007fc88d7dab43 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#12 0x00007fc88d86ca00 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
 ```

The state inside sctp_timeout_handler is 
```
[...]

PR #703, opened 2024-02-29
Assert if soref() is called on an already-released socket.

JonathanLennox:


usrsctplib/user_socketvar.h
```diff
@@ -403,15 +403,18 @@ void	sofree(struct socket *so);
...
 	++(so)->so_count;						\
 } while (0)
 
 #define	sorele(so) do {							\
 	ACCEPT_LOCK_ASSERT();						\
 	SOCK_LOCK_ASSERT(so);						\
 	KASSERT((so)->so_count > 0, ("sorele"));			\
-	if (--(so)->so_count == 0)					\
+	if (--(so)->so_count == 0) {					\
+		(so)->so_count = -1;					\
```
tuexen, line 415: If we actually do this, `sofree()` will return early, since `so->so_count != 0` would be true.

JonathanLennox: Yes, you're right; see my follow-up commit.

JonathanLennox:
The reason for this PR is to try to track down the issue I reported in https://github.com/sctplab/usrsctp/issues/676 -- the garbage value in so_count implies that the memory has been re-used, for which the most likely explanation is that the socket has been freed.

usrsctplib/user_socketvar.h
JonathanLennox: I've rebased it.

PR #707, opened 2024-04-15
Add SCTPDBG statement for connections closing.

JonathanLennox:
This reuses the existing sctp_log_closing mechanism.

### DIFF
--- a/usrsctplib/netinet/sctputil.h
+++ b/usrsctplib/netinet/sctputil.h
@@ -340,10 +340,6 @@ void rto_logging(struct sctp_nets *net, int from);
 
 void sctp_log_closing(struct sctp_inpcb *inp, struct sctp_tcb *stcb, int16_t loc);
 
-#ifdef SCTP_DEBUG
-#define SCTP_LOG_CLOSING 1
-#endif
-
 void sctp_log_lock(struct sctp_inpcb *inp, struct sctp_tcb *stcb, uint8_t from);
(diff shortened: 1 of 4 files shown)


JonathanLennox:
I picked an unused SCTP_DEBUG constant for this; if there's a better value I should use please let me know.

usrsctplib/netinet/sctputil.h
```diff
@@ -340,6 +340,10 @@ void rto_logging(struct sctp_nets *net, int from);
 
 void sctp_log_closing(struct sctp_inpcb *inp, struct sctp_tcb *stcb, int16_t loc);
 
+#ifdef SCTP_DEBUG
+#define SCTP_LOG_CLOSING 1
```
tuexen, line 344: `SCTP_LOG_CLOSING` should be defined via the build system. Since you use it for debugging, maybe something like
`env CFLAGS=-DSCTP_LOG_CLOSING ./configure`
is good enough?

JonathanLennox: I was trying to put the new log statement in a single place, and the existing `sctp_log_closing` function was only called if the `SCTP_LOG_CLOSING` define is set.

Currently (prior to my patch), `sctp_log_closing` only does something if `SCTP_LOCAL_TRACE_BUF` is set.

With my latest commit, the behavior (other than the addition of the new debug message) should be the same as it was previously for all compiler flags.

PR #730, opened 2025-07-21
Fix build with MinGW

tomrittervg:
Tor builds Firefox with WebRTC enabled as of recently, and these fixes are needed to compile usrsctp with MinGW


tomrittervg:
@tuexen Hi, I wanted to ping on this and ask if there was a path forward for this.  Thank you.

tuexen:
Thanks for the patch!

PR #743, opened 2026-05-22
Add LWIP support

vikramdattu:
## Summary

Adds an optional lwIP backend for the userspace socket layer, gated by the new cmake option `sctp_use_lwip` (and the macro `SCTP_USE_LWIP`). When enabled, usrsctp drives sends/receives through lwIP's raw IP/UDP APIs instead of POSIX BSD sockets, so the library can run on bare-metal / RTOS targets that have lwIP but no POSIX socket layer.

Default build path (POSIX) is unchanged: with `sctp_use_lwip=OFF` (the default), no new code is compiled in — no behavioural or ABI delta vs current master.

This is two commits:
1. **Add LWIP support** — the backend itself (25 files, +703/−264).
2. **tests: add host-lwIP smoke test and CI workflow** — proves the new flag works (5 files, +335).

## Context

Supersedes the stale [#583](https://github.com/sctplab/usrsctp/pull/583) by @ycyang1229 (opened 2021-05-03, no upstream activity since). We've been carrying it as a downstream patch in Espressif's `esp_usrsctp` component for production WebRTC use on ESP32-P4 (KVS WebRTC SDK + libsrtp2), and would like to land it upstream so consumers can drop the local patch.

## Changes

- New cmake option `sctp_use_lwip` (default OFF). When ON, `SCTP_USE_LWIP` is defined for the whole tree.
- New BSD-port headers under `usrsctplib/netinet/`: `sctp_in_port.h`, `sctp_ip_port.h`, `sctp_udp_port.h`, `sctp_cc_functions.h` — these isolate the parts of the BSD socket API surface that lwIP replaces.
- Conditional rerouting in `sctp_os_userspace.h`, `sctp_bsd_addr.c`, `sctp_input.c`, `sctp_output.c`, `sctp_pcb.c`, `sctp_timer.c`, `sctp_usrreq.c`, `sctputil.c`, `sctp6_usrreq.c`, `user_socket.c`, `user_recv_thread.c`, `user_environment.c` — all guarded by `#if defined(SCTP_USE_LWIP)` / `#if !defined(SCTP_USE_LWIP)`.
- 22 conditional sites total. No edits outside the lwIP-guarded paths.

## Attribution

The original work is @ycyang1229's. Their PR sat without upstream review and the branch is no longer maintained. We've kept their name as the commit's `Author:` field and added an `Original-Author:` trailer; Vikram Dattu is `Committer` / `Co-authored-by:` since this version was rebased + cleaned by us.

## What changed from #583

- Rebased on current `sctplab/master` (HEAD at submission: `fd070e0 Fix build with MinGW (#730)`). Conflicts resolved via `git am --3way`.
- Folded in the buildability fixes that #583's follow-ups never landed: the `sctp_use_lwip=OFF` default now stays clean, and the IDF Linux-target build path (POSIX) is also fixed. A single toggle of `sctp_use_lwip` produces a clean build both ways.
- Fixed a latent bug in #583's UDP-encapsulation paths in `sctp_output.c`: `GET_UDP_CHKSUM(ip)` was reading/writing the UDP checksum at the IP header's address (silent corruption when SCTP-over-UDP is in use). 16 call sites corrected to `GET_UDP_CHKSUM(udp)`, plus a double-`s_addr` deref on `GET_IP_SRC_ADDR(ip)`. Never bit our WebRTC consumer because we don't use UDP encapsulation, but worth flagging.
- Stripped all platform-specific glue. The upstream version is strictly lwIP-only; anything ESP-IDF-specific (thread-safe netif wrappers, SPIRAM thread stacks, etc.) stays as downstream patches in our component.

## Build verification

POSIX (default `sctp_use_lwip=OFF`):
```
cmake -Dsctp_werror=ON .. && make -j   # passes — 100% Built target usrsctp
```

Downstream / real-world `sctp_use_lwip=ON`: we rebuilt the `streaming_only` example in our ESP32-P4 WebRTC SDK port against this branch — 1903/1903 ninja targets, final binary 2.4 MB (17% partition free), no compile or link errors. This is the same path that has been in production for our pet-camera consumer.

### Host-lwIP smoke test

To let reviewers verify the lwIP path without an ESP-IDF toolchain, the second commit adds `.github/workflows/lwip.yml`:

- **`lwip-options`** (must pass): validates the new `sctp_build_lwip_smoke` cmake option machinery — confirms `FATAL_ERROR` fires when `sctp_use_lwip=OFF`, when `LWIP_SRC_DIR` / `LWIP_PORT_DIR` are unset, and when they point at a non-lwIP directory; then re-runs the default POSIX `make -j` to prove the new option doesn't perturb the existing build path when left OFF.
- **`lwip-smoke`** (`continue-on-error: true` for now): clones lwIP + lwip-contrib at `STABLE-2_2_1_RELEASE`, builds the unix-port archives, configures usrsctp with `-Dsctp_use_lwip=ON -Dsctp_build_lwip_smoke=ON -Dsctp_inet6=OFF`, builds `tests/lwip_smoke.c` (`usrsctp_init` → `usrsctp_socket(AF_INET, SCTP)` → `usrsctp_close` → `usrsctp_finish`), and checks for `lwip_smoke: PASS` on stdout.

The smoke job is best-effort because vanilla lwIP's unix-port + the lwIP shim still have a couple of header-collision rough edges (`struct ip6_hdr` member names, `IP_HDRINCL`) that need a host-side `lwipopts.h` tuning pass — the downstream ESP-IDF build sidesteps these by supplying its own lwipopts. The downstream consumer build above is the load-bearing proof that the backend works in real use; the smoke job is a starting point we'd happily flip to a hard gate once the host lwipopts tuning lands. None of that affects POSIX builds or the `sctp_use_lwip=OFF` default.

Also folded in: guarded the `SCTP_USE_LWIP` `CMSG_ALIGN` fallback in `sctp_os_userspace.h` with `#ifndef CMSG_ALIGN` (Linux glibc / macOS supply it, so the previous unconditional `#define` tripped `-Werror=macro-redefined` on the smoke build).

Happy to split this further, rename the option, or rework anything that doesn't fit. Reviews welcome.

### DIFF
--- a/.github/workflows/lwip.yml
+++ b/.github/workflows/lwip.yml
@@ -1,0 +1,177 @@
+name: Build with lwIP backend
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  lwip-options:
+    # Validates the sctp_build_lwip_smoke option machinery (FATAL_ERROR
+    # paths and the basic configure step).  Always runs to completion
+    # and must pass.
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout usrsctp
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            build-essential cmake git ca-certificates
+
+      - name: Configure rejects sctp_build_lwip_smoke=ON without sctp_use_lwip
+        run: |
+          set +e
+          mkdir -p /tmp/cfg-noLwip
+          cd /tmp/cfg-noLwip
+          OUT="$(cmake "$GITHUB_WORKSPACE" -Dsctp_build_lwip_smoke=ON 2>&1)"
+          rc=$?
+          echo "$OUT"
+          if [ "$rc" -eq 0 ]; then
+            echo "::error::configure should have FATAL_ERROR'd"
+            exit 1
+          fi
+          echo "$OUT" | grep -q 'sctp_build_lwip_smoke=ON requires sctp_use_lwip=ON'
+
+      - name: Configure rejects bogus LWIP_SRC_DIR
+        run: |
+          set +e
+          mkdir -p /tmp/cfg-bogus
+          cd /tmp/cfg-bogus
+          OUT="$(cmake "$GITHUB_WORKSPACE" \
+              -Dsctp_use_lwip=ON \
+              -Dsctp_build_lwip_smoke=ON \
+              -DLWIP_SRC_DIR=/tmp/bogus \
+              -DLWIP_PORT_DIR=/tmp/bogus 2>&1)"
+          rc=$?
+          echo "$OUT"
+          if [ "$rc" -eq 0 ]; then
+            echo "::error::configure should have FATAL_ERROR'd"
+            exit 1
+          fi
+          echo "$OUT" | grep -q "does not look like a lwIP source tree"
+
+      - name: Configure with sctp_build_lwip_smoke=OFF (default path)
+        run: |
+          mkdir -p /tmp/cfg-default
+          cd /tmp/cfg-default
+          cmake "$GITHUB_WORKSPACE"
+          # Default POSIX build must still compile to prove the new
+          # option doesn't perturb anything when left OFF.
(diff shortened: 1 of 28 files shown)


vikramdattu:
@tuexen @tomrittervg @trofi PTAL

